PR #90, opened 2021-02-22
[ATS-113] Add possibility to replay route from S3File.

robertkuraj:
Create S3-files list ViewController with the possibility to replay chosen route.


PublisherExample/amplifyconfiguration.json
```diff
@@ -0,0 +1 @@
+
```
QuintinWillison, line 1: I'm not aware of any benefit of keeping an empty placeholder here. This file should probably be removed from the repository, especially considering it's explicitly Git ignored.

robertkuraj: You are right. I created #114 for this.

PR #93, opened 2021-02-24
[ATS-244] Send elevation and battery status in location update message.

robertkuraj:


QuintinWillison:
We need to get thru the PR backlog in this repository, so am going to land this now, but would still appreciate it if you could give it a once-over if you get a chance @kavalerov. Thanks.

PR #98, opened 2021-03-11
[ATS-120] Update example apps.

robertkuraj:


QuintinWillison (approved):
Given this is just modifying the example apps then I've given it a very cursory review.

PR #104, opened 2021-03-16
Change the way of sending Accuracy enum value. 

### LINKED ISSUE #99: Unify the way of parsing enums sent over Ably channels across all platforms
We need to unify the way of parsing enums because now it's different on each platform. We've agreed to send enum values as strings written in `SCREAMING_CAMEL_CASE`. For example, if the enum has the value `HighAccuracy` it sent as `HIGH_ACCURACY`.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-747) by [Unito](https://www.unito.io)

robertkuraj:
Change the way of sending Accuracy value from sending `Int` to `SCREAMING_CAMEL_CASE`ed `String`.

Old example:
`"accuracy": 1`

New example:
`"accuracy": "MINIMUM_ACCURACY"`

Closes #99.

### DIFF
--- a/Core/Sources/Model/ErrorInformation.swift
+++ b/Core/Sources/Model/ErrorInformation.swift
@@ -31,6 +31,11 @@ public enum ErrorInformationType {
     case trackableAlreadyExist(trackableId: String)
     
     /**
+     Thrown in case when Publisher has already stopped.
+     */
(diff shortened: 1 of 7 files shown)


ricardopereira (approved):
LGTM

PR #124, opened 2021-03-25
[115] Check if client connection state has changed.

### LINKED ISSUE #115: The trackable state should be updated in the Publisher only after the state is different from the previous one.
On iOS, we're always emitting the new trackable state when something happened that triggered state calculation, even if the new state is the same as it was before. 
On Android, we're only emitting a new state if it's different from the previous one (i.e. if the state went from Offline to Online). 
Should we change the behaviour on any of the platforms or are both of those approaches acceptable?

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ATS-308) by [Unito](https://www.unito.io/learn-more)
┆Issue Number: ATS-308

robertkuraj:
Closes #115.

### DIFF
--- a/Publisher/Sources/Services/DefaultAblyPublisherService.swift
+++ b/Publisher/Sources/Services/DefaultAblyPublisherService.swift
@@ -5,6 +5,7 @@ class DefaultAblyPublisherService: AblyPublisherService {
     private let client: ARTRealtime
     private let presenceData: PresenceData
     private var channels: [Trackable: ARTRealtimeChannel]
+    private var clientConnectionState: ConnectionState = .offline
 
     weak var delegate: AblyPublisherServiceDelegate?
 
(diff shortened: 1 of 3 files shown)


Publisher/Sources/Services/DefaultAblyPublisherService.swift
```diff
@@ -18,13 +18,15 @@ class DefaultAblyPublisherService: AblyPublisherService {
 
     private func setup() {
         client.connection.on { [weak self] stateChange in
-            guard let current = stateChange?.current,
-                  let self = self
-            else { return }
-            logger.debug("Connection to Ably changed. New state: \(current)", source: "DefaultAblyPublisherService")
+            guard let self = self,
+                  let receivedConnectionState = stateChange?.current.toConnectionState() else {
```
ricardopereira, line 22: `toConnectionState` is a method from an extension? It converts Ably connection state to the project state? Is that it?

Publisher/Sources/Services/DefaultAblyPublisherService.swift
```diff
@@ -18,13 +18,15 @@ class DefaultAblyPublisherService: AblyPublisherService {
...
+                return
+            }
+            
+            logger.debug("Connection to Ably changed. New state: \(receivedConnectionState)", source: "DefaultAblyPublisherService")
```
ricardopereira, line 26: > `source: "DefaultAblyPublisherService"`

I'm wondering if there's any `#class` literal like `#function`, `#file`, etc. If so, then I think we can use it instead of the literal string, it's safer. I usually forgot to change strings like this when the class is renamed 😄 

Maybe something like `String(DefaultAblyPublisherService.self)` works too, I guess there's something to get the name of the class.

EDIT: or extracting the name from the `#file` literal.

ricardopereira (commented):
I left some comments but it looks good.

Publisher/Sources/Services/DefaultAblyPublisherService.swift
robertkuraj: Yes, you are right. 

```
extension ARTRealtimeConnectionState {
    func toConnectionState() -> ConnectionState {
        ...
    }
}
```

Publisher/Sources/Services/DefaultAblyPublisherService.swift
robertkuraj: I created [issue](https://github.com/ably/ably-asset-tracking-cocoa/issues/127) for this, because it is not the only place where it has to be changed :)

ricardopereira (approved):

PR #129, opened 2021-03-30
[126] Fix problem with sending location update.

### LINKED ISSUE #126: EnhancedLocationUpdateMessage should be sent and received as a single message, not an array.
On Android EnhancedLocationUpdateMessage is being sent and received as a single message.
On iOS it is being sent and received as an array of messages.

iOS need to use the same approach as on Android.

robertkuraj:
Closes #126.


QuintinWillison (approved):
It looks like more work has gone into this pull request than was purely described by the issue it'll close (#126) - e.g. enum alignment - so @robertkuraj, you may want to double check if there are any other issues that could have been linked to be closed by this.

PR #137, opened 2021-05-05
Conform JSON representation of presence message client type values

### LINKED ISSUE #136: Change names of client type values from presence data
We've overlooked one place when we were applying our naming convention for enum values https://github.com/ably/ably-asset-tracking-android/issues/207.

The client types sent in the `PresenceData` are currently written in lowercase `"subscriber", "publisher"`. We should change it to `"SUBSCRIBER", "PUBLISHER"`.

We should be extra careful with this change as **it will break the compatibility with older SDK versions as well as other platforms** (the other platforms will start working as soon as they make this change on their sides too).

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ATS-361) by [Unito](https://www.unito.io/learn-more)
┆Issue Number: ATS-361

QuintinWillison:
Fixes #136.

### DIFF
--- a/Core/Sources/Model/PresenceData.swift
+++ b/Core/Sources/Model/PresenceData.swift
@@ -1,8 +1,8 @@
 import UIKit
 
 public enum ClientType: String, Codable {
-    case publisher
-    case subscriber
+    case publisher = "PUBLISHER"
+    case subscriber = "SUBSCRIBER"
 }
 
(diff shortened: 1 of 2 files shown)


owenpearson (approved):
LGTM, thanks

KacperKluka (approved):
LGTM :+1:

PR #144, opened 2021-05-21
`tracking` namespace added to each channel used for tracking

ben-xD:
I guess this is why this ticket exists, but I might be wrong:
Currently, users need to add the persistence channel rule for each tracking ID in the Ably.com application dashboard. With this, they only need to add the persistence channel rule to the `tracking` namespace once on Ably.com.

Very similar to https://github.com/ably/ably-asset-tracking-cocoa/pull/143 but I used extensions.

Please also let me know if this should be tested.

Closes https://github.com/ably/ably-asset-tracking-cocoa/issues/139

### DIFF
--- a/Core/Sources/AblyExtensions/ARTRealtimeChannels+Extensions.swift
+++ b/Core/Sources/AblyExtensions/ARTRealtimeChannels+Extensions.swift
@@ -3,7 +3,7 @@ import Ably
 private let channelNamePrefix = "tracking"
 
 extension ARTRealtimeChannels {
-    public func getChannelFor(trackingId: String, options: ARTRealtimeChannelOptions? = nil) -> ARTRealtimeChannel {
+    func getChannelFor(trackingId: String, options: ARTRealtimeChannelOptions? = nil) -> ARTRealtimeChannel {
         let segments = [channelNamePrefix, trackingId]
         let channelName = segments.joined(separator: ":")
         if let options = options {
--- a/Core/Sources/AblyExtensions/ARTRealtimeChannels+Extensions.swift
+++ b/Core/Sources/AblyExtensions/ARTRealtimeChannels+Extensions.swift
@@ -1,10 +1,10 @@
 import Ably
 
-let CHANNEL_NAME_PREFIX = "tracking"
+private let channelNamePrefix = "tracking"
 
 extension ARTRealtimeChannels {
     public func getChannelFor(trackingId: String, options: ARTRealtimeChannelOptions? = nil) -> ARTRealtimeChannel {
-        let segments = [CHANNEL_NAME_PREFIX, trackingId]
+        let segments = [channelNamePrefix, trackingId]
         let channelName = segments.joined(separator: ":")
         if let options = options {
             return self.get(channelName, options: options);
--- a/Core/Sources/AblyExtensions/ARTRealtimeChannels+Extensions.swift
+++ b/Core/Sources/AblyExtensions/ARTRealtimeChannels+Extensions.swift
@@ -7,7 +7,7 @@ extension ARTRealtimeChannels {
         let segments = [trackingChannelNamespace, trackingId]
         let channelName = segments.joined(separator: ":")
         if let options = options {
-            return self.get(channelName, options: options);
+            return self.get(channelName, options: options)
         }
         return self.get(channelName)
     }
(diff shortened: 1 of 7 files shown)


paddybyers:
> Currently, users need to add the persistence channel rule for each tracking ID in the Ably.com application dashboard. With this, they only need to add the persistence channel rule to the `tracking` namespace once on Ably.com

This exists so that:
1) channels corresponding to trackingIds are a specific subset of channels in an app. This in the future allows us, or applications, to use other channels in an app for other use-cases without polluting the space of trackingIds;

2) certain attributes that might be required for tracking channels, such as `persistence`, can be configured by namespace.

So yes, the example you give is one of the reasons.

> users need to add the persistence channel rule for each tracking ID in the Ably.com application dashboard

In practice that wouldn't be possible. Creating a new `trackingId` has to be possible without interacting with the dashboard (or equivalent API)

Core/Sources/AblyExtensions/ARTRealtimeChannels+Extensions.swift
```diff
@@ -0,0 +1,14 @@
+import Ably
+
+private let channelNamePrefix = "tracking"
+
+extension ARTRealtimeChannels {
+    func getChannelFor(trackingId: String, options: ARTRealtimeChannelOptions? = nil) -> ARTRealtimeChannel {
+        let segments = [channelNamePrefix, trackingId]
+        let channelName = segments.joined(separator: ":")
```
KacperKluka, line 8: Maybe we could simplify this to `let channelName = "tracking:" + trackingId` or something more Swift idiomatic but equally straightforward? :thinking: It's much easier to understand and reason about :wink:

Core/Sources/AblyExtensions/ARTRealtimeChannels+Extensions.swift
```diff
@@ -0,0 +1,14 @@
+import Ably
+
+private let channelNamePrefix = "tracking"
+
+extension ARTRealtimeChannels {
+    func getChannelFor(trackingId: String, options: ARTRealtimeChannelOptions? = nil) -> ARTRealtimeChannel {
```
KacperKluka, line 6: Is this extension accessible outside of our SDKs? I guess it shouldn't be visible for our SDK users but I don't know enough of iOS and Swift to tell if this is how it works now :wink:

KacperKluka (commented):
Overall looks good but I have a few questions :wink:

Core/Sources/AblyExtensions/ARTRealtimeChannels+Extensions.swift
ben-xD: I think this is quite debatable. I did your idea initially but I moved to my current implementation because this is more extensible with little cost to readability:
```swift
        // your suggestion would look like this:
        let channelName = "\(trackingChannelNamespace):\(trackingId)" // or
        let channelName = trackingChannelNamespace + ":" + trackingId

        // my new implementation:
        let segments = [trackingChannelNamespace, trackingId]
        let channelName = segments.joined(separator: ":")
``` 

If we were to add more segments to the channel ID, mine would be just 1 variable in the segments. It can also be forgotted, imagine someone doing `trackingChannelNamespace + ":" + capabilityScope + trackingId`, and forgetting the `:`. This avoids that, and is quite rigid/ explicit that its joining the segments with colons.

KacperKluka: I agree with all the benefits that your solution brings but the question is if we are going to use them? (are we going to change this code in the near future?) :thinking: In such cases, I like to go with the YAGNI rule and assume that we won't and additionally try to stick to KISS rule and make it as simple as possible :wink: But it all boils down to personal preference in this case so I'm happy to go with your solution if that's what you prefer :smile:

Core/Sources/AblyExtensions/ARTRealtimeChannels+Extensions.swift
ben-xD: Thanks for mentioning this, currently its following the same style as another extension:
```swift
extension ARTRealtimeChannelState {
    func toConnectionState() -> ConnectionState {
        switch self {
        case .attached:
            return .online
        case .failed:
            return .failed
        default:
            return .offline
        }
    }
}
```
Because no access modifier is specified, it is by default **internal access** only.  If it is meant to public, you will should mark it as open or public. So this won't be accessible to other users, I confirmed it with: 

Screenshot 2021-05-24 at 13 46 01

KacperKluka: Awesome, thanks for checking :+1:

QuintinWillison (approved):

KacperKluka (approved):

PR #146, opened 2021-05-24
Make battery level optional in EnhancedLocationUpdateMessage

ben-xD:
This allows `EnhancedLocationUpdateMessage` to be optional, to receive messages from other libraries (e.g. Android) without crashing.

From @KacperKluka in Slack messages: https://ably-real-time.slack.com/archives/C01EPJENRD0/p1621852415046500:
> In the `EnhancedLocationUpdateMessage` the batteryLevel is just a Float which means it doesn't accept null. For parsing JSON we're using JSONDecoder.decode() which throws an error if the data isn't of the expected type. This means that if in the JSON message the `batteryLevel` will be null or missing then the decoder will fail. Luckily, the app won't crash in this case as we're handling such error in the code. However, all location updates without battery level will get lost making the iOS subscriber useless if the publisher is started on Android :confused: To fix it we should change the batter level type to `Float?` which accepts nulls.

An example error when decoding a JSON string:
```
**Swift.DecodingError.valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "first", intValue: nil)], debugDescription: "Expected String value but found null instead.", underlyingError: nil)**
```

### DIFF
--- a/Core/Sources/Model/LocationUpdate.swift
+++ b/Core/Sources/Model/LocationUpdate.swift
@@ -25,13 +25,13 @@ public class EnhancedLocationUpdate {
 
 public class EnhancedLocationUpdateMessage: Codable {
     let location: GeoJSONMessage
-    let batteryLevel: Float
+    let batteryLevel: Float?
     let intermediateLocations: [GeoJSONMessage]
     let type: LocationUpdateType
 
     init(locationUpdate: EnhancedLocationUpdate, batteryLevel: Float?) throws {
         self.location = try GeoJSONMessage(location: locationUpdate.location)
-        self.batteryLevel = batteryLevel ?? 0
+        self.batteryLevel = batteryLevel
         self.intermediateLocations = []
         self.type = locationUpdate.type
     }
--- a/Core/Tests/Sources/Model/EnhancedLocationUpdateTests.swift
+++ b/Core/Tests/Sources/Model/EnhancedLocationUpdateTests.swift
@@ -1,0 +1,22 @@
+import Foundation
+import CoreLocation
+import XCTest
+@testable import Core
+
+class EnhancedLocationUpdateTests: XCTestCase {
+
+    func testEnhancedLocationUpdateMessageDecodesNullBatteryLevel() throws {
+        let location = CLLocation(latitude: 0.0, longitude: 0.0)
+        let expected = try! EnhancedLocationUpdateMessage(locationUpdate: EnhancedLocationUpdate(location: location), batteryLevel: nil)
+        
+        let jsonString = try! expected.toJSONString()
+        let actual: EnhancedLocationUpdateMessage = try! EnhancedLocationUpdateMessage.fromJSONString(jsonString)
+        
+        XCTAssertEqual(expected.batteryLevel, actual.batteryLevel)
+        XCTAssertEqual(expected.type, actual.type)
+        
+// There are more things to XCTAssertEqual but you'll need to implement Equatable or start using Structs instead of classes.
+//        XCTAssertEqual(expected.location, actual.location)
+//        XCTAssertEqual(expected.intermediateLocations, expected.intermediateLocations)
+    }
+}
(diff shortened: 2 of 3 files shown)


Core/Sources/Model/LocationUpdate.swift
```diff
@@ -25,13 +25,13 @@ public class EnhancedLocationUpdate {
 
 public class EnhancedLocationUpdateMessage: Codable {
```
KacperKluka, line 26: This is only used in the internal communication between Subscriber and Publisher, right? It's not visible from the outside of our SDKs? I'm asking because I've noticed that it's a `public class` :wink:

Core/Tests/Sources/Model/EnhancedLocationUpdateTests.swift
```diff
@@ -0,0 +1,22 @@
...
+        
+        let jsonString = try! expected.toJSONString()
+        let actual: EnhancedLocationUpdateMessage = try! EnhancedLocationUpdateMessage.fromJSONString(jsonString)
+        
+        XCTAssertEqual(expected.batteryLevel, actual.batteryLevel)
+        XCTAssertEqual(expected.type, actual.type)
+        
+// There are more things to XCTAssertEqual but you'll need to implement Equatable or start using Structs instead of classes.
+//        XCTAssertEqual(expected.location, actual.location)
+//        XCTAssertEqual(expected.intermediateLocations, expected.intermediateLocations)
```
KacperKluka, line 20: I'm wondering if we should have those other assertions here as the purpose of this test is to check the `batteryLevel` field :thinking: But I'll leave it up to you :wink:

Core/Tests/Sources/Model/EnhancedLocationUpdateTests.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+import CoreLocation
+import XCTest
+@testable import Core
+
+class EnhancedLocationUpdateTests: XCTestCase {
+
+    func testEnhancedLocationUpdateMessageDecodesNullBatteryLevel() throws {
```
KacperKluka, line 8: After merging that Android PR https://github.com/ably/ably-asset-tracking-android/pull/343 the battery level will be missing from the message JSON so maybe let's create a test for the scenario in which there is no battery level in the JSON? :thinking:

KacperKluka (commented):
Overall looks good :+1: I have a few questions though :wink: 

@QuintinWillison I'm wondering if we should maybe just remove the battery level completely instead of just making it nullable since soon the [Android PR](https://github.com/ably/ably-asset-tracking-android/pull/343) that does that will be landed? :thinking:

Core/Sources/Model/LocationUpdate.swift
ben-xD: Thanks for catching that. I left it the way it was originally and didn't think about it. I noticed that many classes which seem to be for internal use are public in other files too. I've removed public from the other classes in the file (`LocationUpdateType`, `EnhancedLocationUpdate`, and `EnhancedLocationUpdateMessage`), but there are many more instances of over-exposure 😅.

Core/Tests/Sources/Model/EnhancedLocationUpdateTests.swift
ben-xD: I've deleted this test because by removing the battery level completely, there is nothing to test. Let me know if you think there are some tests I should write, thanks :)

Core/Tests/Sources/Model/EnhancedLocationUpdateTests.swift
ben-xD: Same reply as https://github.com/ably/ably-asset-tracking-cocoa/pull/146#discussion_r638056120

ben-xD:
@KacperKluka after you suggested just removing battery level, there's been some changes to the PR.

PR #147, opened 2021-05-24
Remove battery level from `EnhancedLocationUpdateMessage`

ben-xD:
Initially I was going to make battery level optional, but there was another ticket that just removes it because it wasn't being used anyway: 

So I have closed the initial PR https://github.com/ably/ably-asset-tracking-cocoa/pull/146 in favour of this one.

Closes https://github.com/ably/ably-asset-tracking-cocoa/issues/140


Core/Sources/Model/LocationUpdate.swift
```diff
@@ -3,15 +3,15 @@ import CoreLocation
...
+enum LocationUpdateType: String, Codable {
     case predicted = "PREDICTED"
     case actual = "ACTUAL"
 }
 
 /**
  Model used to handle location updates.
  */
-public class EnhancedLocationUpdate {
+class EnhancedLocationUpdate {
```
KacperKluka, line 14: This should still be `public` because it's a class that's exposed in the API :wink: Only the classes that are used for Publisher-Subscriber communication via Ably channels should be private. I'm a bit surprised that after making it invisible outside the SDK the checks have passed :thinking: Aren't you using it in your example apps code?

Core/Sources/Model/LocationUpdate.swift
```diff
@@ -3,15 +3,15 @@ import CoreLocation
 /**
  Enumeration used to determine enhanced location type.
  */
-public enum LocationUpdateType: String, Codable {
+enum LocationUpdateType: String, Codable {
```
KacperKluka, line 6: It should remain `public` just like `EnhancedLocationUpdate` :wink:

KacperKluka (changes requested):
Looks good but you've made too many classes private :smile:

Core/Sources/Model/LocationUpdate.swift
ben-xD: I checked, and the `EnhancedLocationUpdate` and `EnhancedLocationUpdate` is not being used in the example apps.

Okay I think the main issue with setting these public/ not-public is I don't know what the public API is, and am only guessing (quite badly 😅)

Core/Sources/Model/LocationUpdate.swift
ben-xD: Cool, reverted it 👍

KacperKluka (approved):
LGTM :+1:

QuintinWillison (approved):

ben-xD:
I'm not sure if this needs coordination before merging, @KacperKluka I think you know best when to merge, feel free to click merge when ready

KacperKluka:
@ben-xD I think that your PR should be landed before https://github.com/ably/ably-asset-tracking-android/pull/343 since on Android everything will work even if the `batteryLevel` is missing :wink:

PR #152, opened 2021-05-26
Update to Mapbox navigation 2.0 beta 10

ben-xD:
Updated to a beta version of Mapbox navigation.

I also noticed M1 mac workaround is no longer needed. So I removed the code used in Cocoapods related files which did this. I tested that it works on arm64 (m1 mac) and x86_64 (intel) architectures.

This is mainly a PR to remove the M1 mac workaround (and warnings), but I also had the mapbox update working yesterday when I was evaluating how hard it would be to update. It seemed quite easy so here it is too.

Closes https://github.com/ably/ably-asset-tracking-cocoa/issues/106


ben-xD:
You might see that I removed the error handling, but this is because no error is thrown anymore. It's a little hard to find the response from MapBox, but here it is: 
Screenshot 2021-05-26 at 15 25 54

QuintinWillison (approved):

PR #156, opened 2021-05-28
Swift package manager support

ben-xD:
This PR adds support for Swift Package Manager.

It changes the repo from having multiple Xcode projects into 1 Xcode workspace which contains
- 1 Swift package which contains 2 libraries (what the user selects to add into their xcode project/ swift package) and 4 modules/ targets (what the user selects to `import`/ compile into their code). Configuration is done by `Package.swift`, but Xcode can autogenerate schemes to help with running tests and building targets.
- 4 Example apps in the form of xcode projects

I also removed Gemfile/ Gemfile.lock, and bundler. I didn't go as far as to remove `fastlane`, but that can be done in a follow up github issue, since we are not really making use of it (we are just running `xcodebuild` through it). We can use `makefile` instead, in the future.

TODOs
  - [ ] Wait for ably-cocoa to release SPM support. Currently using my [branch](https://github.com/ably/ably-cocoa/tree/feature/ben-butterworth-temporary-spm) of ably-cocoa.

Closes https://github.com/ably/ably-asset-tracking-cocoa/issues/148


ricardopereira:
I don't know what's the adoption percentage of SPM, but I notice there are more SPM users since Xcode 12. Even so, CocoaPod is still a reliable and the most used package manager in the Cocoa world. In my opinion, Ably should support CocoaPods. until it's viable. It's not a big effort, you only need to keep the pod spec updated. About Ably Cococa lib and SPM, there's some work done to support SPM: https://github.com/ably/ably-cocoa/issues/819.

ben-xD:
Yup, in that case, we should leave a basic version of the Podspec in, but this file will be quite empty, so there is no redundant information. That will allow the package to be released/ used on Cocoapods.

e.g. Compare Swiftlints [Podpsec](https://github.com/realm/SwiftLint/blob/master/SwiftLint.podspec) which has no dependencies or targets (they're listed in `Package.swift`), where as we are listing dependencies and targets in our [Podspec](https://github.com/ably/ably-asset-tracking-cocoa/blob/main/AblyAssetTracking.podspec).

ben-xD:
We should consider delaying **support for Cocoapods** to another issue. This is because supporting both SPM and Cocoapods at the same time will require writing additional podspecs (configuration for a cocoapod). This is because each podspec has its own module/ namespace (e.g. import `Podname`). In other words, there is a one to one mapping from a podspec to an `import AblyAssetTracking`. 

However, in order to support Swift Package Manager, we have created multiple modules (Publisher, Subscriber, Internal and Core), and are importing these modules in the other modules. This means we need to specify 4 Podspec files with the same name, so that these `imports` will resolve/ build. Mapbox does a similar structure, which a relatively complex dependency graph. They have [MapboxNavigation podspec](https://github.com/mapbox/mapbox-navigation-ios/blob/main/MapboxNavigation.podspec), depending on [`MapboxCoreNavigation` podspec](https://github.com/mapbox/mapbox-navigation-ios/blob/main/MapboxCoreNavigation.podspec) depending on [`MapboxNavigationNative ` repo](MapboxNavigationNative) (where the podspec does not seem to be open source)

TLDR: As it currently stands, cocoapods is not supported. It is very easy for users (Xcode 11 or newer) to add this SDK as a Swift Package to their Xcode project or their own Swift package. We can do more work to add Cocoapods support, but it will take some effort.

.github/workflows/assemble.yml
```diff
@@ -10,27 +10,27 @@ jobs:
...
           MAPBOX_ACCESS_TOKEN: ${{ secrets.MAPBOX_ACCESS_TOKEN }}
           ABLY_API_KEY: ${{ secrets.ABLY_API_KEY }}
           AMPLIFY_CONFIGURATION: ${{ secrets.AMPLIFY_CONFIGURATION }}
 
       - name: Build All Projects
         run: |
-          bundle install
-          bundle exec fastlane build_subscriber
-          bundle exec fastlane build_publisher
-          bundle exec fastlane build_example_apps
```
lukasz-szyszkowski, line 36: what's the reason for removing the bundler?

Sources/AblyAssetTrackingCore/Model/ClientType.swift
```diff
@@ -0,0 +1,6 @@
+
+
+public enum ClientType: String, Codable {
+    case publisher = "PUBLISHER"
+    case subscriber = "SUBSCRIBER"
```
lukasz-szyszkowski, line 5: until this is for internal use, it's not a good practice to assign a string name to string type enum, especially uppercased. You can leave it as it is if an uppercased name is required by an external API
```suggestion
    case publisher 
    case subscriber
```

Sources/AblyAssetTrackingCore/Model/EnhancedLocationUpdate.swift
```diff
@@ -0,0 +1,24 @@
+import CoreLocation
+
+/**
+ Enumeration used to determine enhanced location type.
+ */
+public enum LocationUpdateType: String, Codable {
+    case predicted = "PREDICTED"
+    case actual = "ACTUAL"
```
lukasz-szyszkowski, line 8: same as before: until this is for internal use, it's not a good practice to assign a string name to string type enum, especially uppercased. You can leave it as it is if an uppercased name is required by an external API
```suggestion
    case predicted
    case actual
```

Sources/AblyAssetTrackingInternal/EnhancedLocationUpdateMessage.swift
```diff
@@ -0,0 +1,13 @@
+import AblyAssetTrackingCore
+
+public class EnhancedLocationUpdateMessage: Codable {
```
lukasz-szyszkowski, line 3: why this is no struct type? It's recommended type for Codable objects.

Sources/AblyAssetTrackingInternal/Utils/Codable+Extensions.swift
```diff
@@ -1,4 +1,22 @@
...
+        - ErrorInformation of JSONCodingError type when we are not able to create `String` from coded `Data`
+        - Any JSONDecoder error: Thrown by the `JSONEncoder` and just passed forward
+     - Returns: JSON string encoded with UTF-8
+     */
+    func toJSONString() throws -> String {
+        let data = try JSONEncoder().encode(self)
+        if let result =  String(data: data, encoding: .utf8) {
+            return result
+        }
+        throw ErrorInformation(type: .JSONCodingError(for: "\(data.self)"))
```
lukasz-szyszkowski, line 17: in this case, the reason for this exception is not clear. It should inform the user that 
```swift
String(data: data, encoding: .utf8)
```
can't create String from data

Sources/AblyAssetTrackingSubscriber/Exports.swift
```diff
@@ -0,0 +1,3 @@
+// This 
+// There have been proposals to work e.g. `public import AblyAssetTrackingCore` to expose the import to users of this module
+@_exported import AblyAssetTrackingCore
```
lukasz-szyszkowski, line 3: https://forums.swift.org/t/what-does-exported-import-do/35869/4

Tests/InternalTests/ConnectionConfigurationTests.swift
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    func testPerformanceExample() throws {
+        // This is an example of a performance test case.
+        self.measure {
+            // Put the code you want to measure the time of here.
+        }
+    }
+
+}
```
lukasz-szyszkowski, line 25: empty test case? Shouldn't exist.

Tests/SystemTests/PublisherSubscriber.swift
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    func testPerformanceExample() throws {
+        // This is an example of a performance test case.
+        self.measure {
+            // Put the code you want to measure the time of here.
+        }
+    }
+
+}
```
lukasz-szyszkowski, line 25: empty test case - remove it

fastlane/Fastfile
```diff
@@ -1,78 +1,51 @@
...
-    scan(
-      workspace: './AblyAssetTracking.xcworkspace',
-      scheme: 'Subscriber',
-      clean: true,
-      code_coverage: true,
+    xcodebuild(
+      scheme: 'SubscriberTests',
+      configuration: 'Debug',
+      destination: 'platform=iOS Simulator,name=iPhone 12',
+      test: "test"
```
lukasz-szyszkowski, line 38: >Note: xcodebuild is a complex command, so it is recommended to use gym for building your ipa file and scan for testing your app instead.
https://docs.fastlane.tools/actions/xcodebuild/

`Scan` won't work?

fastlane/Fastfile
```diff
@@ -1,78 +1,51 @@
...
-      source_directory: './Subscriber/Sources'
+  end
+
+  desc 'Run Publisher tests'
+  lane :test_publisher do
+    xcodebuild(
+      scheme: 'PublisherTests',
+      configuration: 'Debug',
+      destination: 'platform=iOS Simulator,name=iPhone 12',
+      test: "test"
```
lukasz-szyszkowski, line 48: >Note: xcodebuild is a complex command, so it is recommended to use gym for building your ipa file and scan for testing your app instead.
https://docs.fastlane.tools/actions/xcodebuild/

`Scan` won't work?

lukasz-szyszkowski (changes requested):

fastlane/Fastfile
lukasz-szyszkowski: Did you try this solution? https://uptech.team/blog/swift-package-manager 
**chapter**: `2. To Change The Path Of SPM Dependencies, You Need To Override ClonedSourcePackagesDirPath.`

ben-xD: > EDIT: I guess this is the reason

Yes, fastlane scan does not support Swift Packages.

> Did you try this solution?

I'm not sure what you mean. How is changing SPM dependency path a solution? Do you suggest I cache the SPM packages to improve build times? I don't think scan does that? I think that's could be a good github issue to speed up dependency installation, especially since Mapbox dependencies could reach ~100MB.

lukasz-szyszkowski: I mean if `scan` some support for SPM (in this case `cloned_source_packages_path`) I guess it should support SPM somehow?

Sources/AblyAssetTrackingSubscriber/Exports.swift
ben-xD: Yea, so it is not stable. In the future, in a future version of Swift, it will probably be removed and replaced with `public`, at which point that file should be updated. This is to avoid doing `import AblyAssetTrackingCore` on top of `import AblyAssetTrackingPublisher`. 

Do you think I should not have it at all? LMK. I don't think it's dangerous to keep it.

lukasz-szyszkowski: well, AAT is still in Beta so it's safe for now, but even if `public` will be introduced in a future version of swift, this will require updating Xcode IDE from users. Leave it for now, keep it in mind for future 😅

Sources/AblyAssetTrackingCore/Model/EnhancedLocationUpdate.swift
ben-xD: Same comment as above 😅

Sources/AblyAssetTrackingCore/Model/ClientType.swift
ben-xD: That's an existing piece of code I didn't actually write, unfortunately git is confused. I just moved the file into Core module, because it is being shared between 2 other modules (Publisher and Subscriber). I found it odd as well, so I asked Kacper why this is done, and he said if it were done again, it would not be there.

Sources/AblyAssetTrackingInternal/Utils/Codable+Extensions.swift
ben-xD: This is also code I did not write, but just moved because I needed to move them into a new module.

lukasz-szyszkowski: 😅

Tests/SystemTests/PublisherSubscriber.swift
ben-xD: Different to the other one you saw (connectionConfigurationTests). this is one is there because I needed a test file in `SystemTests` module/ test target to actually run this new test target I created. Its empty currently, but I am using it in the Token Authentication feature I am adding. It will only be temporary for this one PR.

Otherwise I got `xcodebuild: error: The workspace named "AblyAssetTracking" does not contain a scheme named "SystemTests"`, because Xcode doesn't generate a Scheme for a test target without any source files.

However, I have **simplified** it in ce627dcea504709b2a45355510e1cfa2c2992eac

Tests/InternalTests/ConnectionConfigurationTests.swift
ben-xD: Deleted in ce627dcea504709b2a45355510e1cfa2c2992eac

Sources/AblyAssetTrackingInternal/EnhancedLocationUpdateMessage.swift
ben-xD: This is also code I did not write, but just moved because I needed to move them into a new module and also add the new `import`. I guess Github got confused. I really think a lot of classes should be structs in this project, but the main reason it did not use a single struct is because it was trying to support Objective-C, and using `@obj-c` or inheriting from NSObject.

.github/workflows/assemble.yml
ben-xD: Mainly because there's no need for it, it works fine without. I guess the question, what is the point of bundler. (Fastlane says speed improvement, but doesn't any concrete reasons).

I wanted to propose removing fastlane as well, and just using Makefiles and xcodebuild, as we discussed on Friday. I didn't want to make any more drastic changes. 

LMK if you think I should re-add it. I think it would be nice to remove fastlane (I felt like it was a popular idea in the meeting we had on Friday), perhaps even in this PR then?

lukasz-szyszkowski: Using bundler is safer in case of a fastlane version,  because it's using Gemfile.resolved ... but for future use, where we want to remove `fastlane` it's a good idea to remove using bundler for fastlane. LGTM

PR #169, opened 2021-06-23
Add token based authentication

ben-xD:
Adds TokenAuthentication support, including TokenRequest, TokenDetails, TokenString/ JWT. I was able to authenticate using TokenRequest, but haven't confirmed TokenDetails, String or TokenDetails. These are minor additions to TokenRequest, so it's expected to work.


Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
```diff
@@ -4,7 +4,7 @@ import MapboxDirections
 import AblyAssetTrackingCore
 
 class DefaultLocationService: LocationService {
-    private let locationDataSource: PassiveLocationDataSource
+    private let locationDataSource: PassiveLocationManager
```
ben-xD, line 7: We renamed the type, but shouldn't the variable also be renamed to match?

Tests/PublisherTests/DefaultPublisher/DefaultPublisherTests.swift
```diff
@@ -21,17 +25,28 @@ class DefaultPublisherTests: XCTestCase {
...
+        try setUpUsingAPIKeyWithError()
+    }
+    
+    override func tearDownWithError() throws {
+        configuration = nil
+    }
+    
+    func setUpUsingAPIKeyWithError() throws {
+        if (configuration != nil) {
+            throw ClientConfigError.redefinedConnectionConfiguration
```
ben-xD, line 47: This is not that readable. Maybe we should add associated value (string) with message, and throw it together? For example:

`throw ClientConfigError.redefinedConnectionConfiguration(message: "The client connection configuration has already been defined.")`

My error message is only a guess, because I am not sure what it's doing there.

ben-xD: After checking the git blame, turns out I wrote that. 😂  20e6c87300de79241f8239fd0d0fd369b60713d7 renames this error enum to be clearer.

Tests/SystemTests/Utils/JWTHelper.swift
```diff
@@ -0,0 +1,48 @@
...
+    ) -> String? {
+        
+        let keyTokens = Secrets.ablyApiKey.split(separator: ":")
+        let keyName = String(keyTokens[0])
+        var keySecret = String(keyTokens[1])
+        if (invalid) {
+            keySecret = "invalid"
+        }
+
+        var urlComponents = URLComponents(string: "https://echo.ably.io/createJWT")
```
ben-xD, line 28: This is a dangerous request? The key name and secret is shared in the URL.

Tests/SystemTests/PublisherSubscriberSystemTests.swift
```diff
@@ -3,7 +3,7 @@ import AblyAssetTrackingCore
 import AblyAssetTrackingPublisher
 import AblyAssetTrackingSubscriber
 
-class PublisherSubscriber: XCTestCase {
```
ben-xD, line 6: The intention of this test class is for the publisher to publish and the subscriber to subscribe to those location updates. This would be the same as the Android library.

Tests/SystemTests/Utils/AuthHelper.swift
```diff
@@ -0,0 +1,36 @@
...
+            requestCompleted = true
+        }
+        
+        while !requestCompleted {
+            CFRunLoopRunInMode(CFRunLoopMode.defaultMode, CFTimeInterval(0.1), false)
+        }
+        
+        return fetchedTokenDetails
+    }
+}
```
ben-xD, line 36: This is an interesting helper method. Would this be a good idea for applications (non test code?)

Tests/SystemTests/Utils/URLSessionServerTrustSync.swift
```diff
@@ -0,0 +1,59 @@
...
+//  File.swift
+//  
+//
+//  Created by Łukasz Szyszkowski on 16/08/2021.
+//
+
+import Foundation
+import XCTest
+
+class URLSessionServerTrustSync: NSObject, URLSessionDelegate, URLSessionTaskDelegate {
```
ben-xD, line 11: Could you please add some comments about what this does. The name is not so friendly. Is the class being used anywhere, I can't find the usage?

Tests/SystemTests/PublisherAuthenticationSystemTests.swift
```diff
@@ -0,0 +1,173 @@
...
+            .connection(configuration)
+            .mapboxConfiguration(MapboxConfiguration(mapboxKey: Secrets.mapboxAccessToken))
+            .log(logConfiguration)
+            .locationSource(LocationSource(locationSource: [CLLocation(latitude: 0.0, longitude: 0.0)]))
+            .routingProfile(.driving)
+            .delegate(publisherDelegate)
+            .resolutionPolicyFactory(DefaultResolutionPolicyFactory(defaultResolution: resolution))
+            .start() // Doesn't start publishing, its just a `build()` publisher call.
+
+        // TODO check that connection is made/ Await successfully connection callback with an expectation
```
ben-xD, line 130: What do you think about this issue @lukasz-szyszkowski ?

Tests/SystemTests/Utils/URLSessionServerTrustSync.swift
ben-xD: For example, with documentation I would know that the sync is for **synchronous**, not **synchronize**. These are utility methods for being able to return values instead of using callbacks. 

I still can't find the usage of them though.

Tests/SystemTests/PublisherAuthenticationSystemTests.swift
lukasz-szyszkowski: I've talked to @KacperKluka some time ago and there is no reason to check if the publisher connects successfully or not, all events are enqueued and will execute asap. But for the testing purposes, it'll be helpful, I agree.

Tests/SystemTests/Utils/URLSessionServerTrustSync.swift
lukasz-szyszkowski: Ok, I'll add comments, this is a copy of helper from `ably-cocoa`

Tests/SystemTests/Utils/JWTHelper.swift
lukasz-szyszkowski: agree, I've copied it from `ably-cocoa` and I assume it should be hidden there too.

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
lukasz-szyszkowski: 9735a96153647b964a04ea66863d2c9403a799d5

Tests/SystemTests/Utils/JWTHelper.swift
ben-xD: I mean its dangerous because the credentials are in the URL. The concern similar to this stack overflow answer: [Passing username and password in HTTP GET query parameters](https://stackoverflow.com/a/36840297/7365866).

> - URLs are stored in web server logs
> - URLs are stored in the browser history
> -URLs are passed in Referrer headers

lukasz-szyszkowski: Ok, but it's not mine. I've copied it from `ably-cocoa` so someone before me "invented" this 🤷‍♂️

Tests/SystemTests/PublisherSubscriberSystemTests.swift
lukasz-szyszkowski: ead77e36171b7a31fd3d07a9077325eee2fe69b7

Tests/SystemTests/Utils/AuthHelper.swift
lukasz-szyszkowski: What did you find interesting here? :)

Tests/SystemTests/Utils/URLSessionServerTrustSync.swift
lukasz-szyszkowski: 1565f2057cdf178e6e1bc1ab61dc90a3b06520b6

Examples/PublisherExample/Sources/Screens/Map/MapViewController.swift
```diff
@@ -81,33 +81,51 @@ class MapViewController: UIViewController {
...
     private func setupPublisher() {
         let resolution = Resolution(accuracy: .balanced, desiredInterval: 5000, minimumDisplacement: 100)
         currentResolution = resolution
-        
+
+        // Authentication to Ably with a private Ably API key
+        let connectionConfiguration = ConnectionConfiguration(apiKey: Environment.ABLY_API_KEY, clientId: "Asset Tracking Publisher Example")
+
+        // Authentication with Ably with an AuthCallback, which is passed a TokenRequest.
+//        let connectionConfiguration = ConnectionConfiguration(clientId: "Asset Tracking Publisher Example") { tokenParams, resultHandler in
```
QuintinWillison, line 98: We need to find a way to ensure this code is compiler visible. Perhaps an enum and switch?

fastlane/testfile.swift
```diff
@@ -0,0 +1 @@
+struct TestFile {}
```
QuintinWillison, line 1: What's this for? Could we have a comment at the start of this file to explain?

Tests/SystemTests/Utils/RestHelper.swift
```diff
@@ -0,0 +1,28 @@
+//
+//  File.swift
```
QuintinWillison, line 2: In this and other new files there is this boilerplate. Please remove these. We can find out who created it from Git history and the boilerplate is misleading / adds cognitive load anyway for reviewers.

Tests/SystemTests/Utils/JWTHelper.swift
QuintinWillison: It's `https` so by the point the query portion has been transmitted we'll be over TLS (see [this answer](https://stackoverflow.com/a/2629241)). Therefore this is not dangerous, given our context.

QuintinWillison (changes requested):

Tests/SystemTests/Utils/JWTHelper.swift
ben-xD: My concern is if for some reason HTTP URLs used were shown in the github workflow (e.g. running in verbose mode to debug an error happening only in the CI), the secrets are leaked in the logs.

lukasz-szyszkowski: oh, who can view actions logs? It's public or restricted access?

Tests/SystemTests/Utils/RestHelper.swift
lukasz-szyszkowski: f5eab85276b7643cc8eb8f5d791e4e24a74d37cf

fastlane/testfile.swift
lukasz-szyszkowski: removed here 043ccb3d0683ec005a700b09ae205021866998de

Examples/PublisherExample/Sources/Screens/Map/MapViewController.swift
lukasz-szyszkowski: ok, as far as I understood ... you want to give users an option to select which type of auth they want to use?

Tests/SystemTests/PublisherAuthenticationSystemTests.swift
```diff
@@ -0,0 +1,171 @@
+import XCTest
+import AblyAssetTrackingCore
+import AblyAssetTrackingPublisher
+import AblyAssetTrackingInternal
+import CoreLocation
+import Ably
+
+class PublisherAuthenticationSystemTests: XCTestCase {
```
KacperKluka, line 8: Is there a reason to have separate tests for Publisher and Subscriber authentication? It introduces code duplication as both those classes are very similar :wink: Couldn't you merge them and create tests that check the authentication methods both on publisher and subscriber at the same time? We did something like that on [Android](https://github.com/ably/ably-asset-tracking-android/blob/main/integration-testing-app/src/androidTest/java/com/ably/tracking/tests/AuthenticationTests.kt)

KacperKluka (approved):
Given my little iOS knowledge it looks good :+1: I'll only leave a small refactor suggestion :wink:

Examples/PublisherExample/Sources/Screens/Map/MapViewController.swift
ben-xD: This is a class I designed. I am confused by this statement too. Maybe you didn't spot that this file is part of the **example** app? looking at `ConnectionConfiguration`, if a user wants to create a `ConnectionConfiguration`, there are multiple convenience initializers to choose on based on their authentication method.

Tests/SystemTests/Utils/JWTHelper.swift
ben-xD: The log access on Github turns out to be restricted from someone without a Github account. I am not sure of the full access control or this, or how to control this. What if we want to make it public so contributors/ external parties can see the logs? That same answer Quintin shared has my concern: 

> Yes. The querystring is also encrypted with SSL. Nevertheless, as this article shows, it isn't a good idea to put sensitive information in the URL. For example: ...

https://stackoverflow.com/a/2629241

Tests/SystemTests/Utils/AuthHelper.swift
ben-xD: Mostly `CFRunLoopRunInMode`, I was wondering when you should use this? And if there are other options here, e.g. semaphore's, which are more conventionally used. https://stackoverflow.com/a/40491615/7365866

Tests/SystemTests/Utils/URLSessionServerTrustSync.swift
ben-xD: Can you show me where this class is being used? I think you might have missed when I said this at first

> Is the class being used anywhere, I can't find the usage?

ben-xD: Okay, got it, its in `JWTHelper.swift`, 
```
let (responseData, responseError, _) = URLSessionServerTrustSync().get(request)
```

Tests/SystemTests/Utils/JWTHelper.swift
lukasz-szyszkowski: ok, 2 questions:
- is it a blocker for this PR?
- do we have an endpoint where we can send data using POST method?

Tests/SystemTests/Utils/AuthHelper.swift
lukasz-szyszkowski: This code is copy-paste from `ably-cocoa` SDK.

We can't use a semaphore here, because the intention is to block the main thread https://stackoverflow.com/a/54104259
I guess `CFRunLoopRunInMode` config is just for slowing down run loop

ben-xD: > This code is copy-paste from ably-cocoa SDK.

Okay 😂. I'm not sure what you mean by **slowing down run loop**. I'm trying to understand that line but I don't have any understanding at the moment.

Do you know what `CFRunLoopRunInMode` **does**? It looks like it's used to block the main thread, just like a semaphore would. I think a semaphore would be more obvious

Tests/SystemTests/Utils/JWTHelper.swift
ben-xD: I've just checked, these logs are visible for anyone with Github account, so it is **public**. If logs are printed, these secrets will get leaked.

> do we have an endpoint where we can send data using POST method?

The endpoint you're using isn't even documented, so I wouldn't expect POST to be documented 😅

At the end of the day, it is the API design which asks for key information as query param, so the ship has sailed on fixing this. I am looking for a POST method. I think it depends on code quality requirements. 

Anyway, please **resolve** this once you read this message. 😂

lmars: I've just come to this conversation from a question asked in Slack, so forgive me if I'm missing prior context, but with respect to this point:

> The secret key and the key name are injected from github.SECRETS in CI

Is there a reason we're not using a temporary Sandbox app created by the tests themselves, rather than a long lived API key in GitHub secrets? This reduces the impact of the key leaking, since it is short lived and has low limits.

See the G2 testing guideline: https://docs.ably.io/client-lib-development-guide/features/#G2

Also, as mentioned to @ben-xD in Slack, this endpoint isn't publicly documented, but exists in a private repository that soon may become public: https://github.com/ably/echoserver/blob/master/app.js#L162-L202

It does accept POST requests, and here are the fields the body can contain:

```
/**
 * Handles GET and POST requests. Can accept the following params:
 *
 *  - keyName:     the key name.
 *  - keySecret:   the key secret.
 *  - expiresIn:   the token duration. Represents seconds. Optional.
 *  - clientId:    the clientId. Optional.
 *  - jwtType:     use 'embedded' to obtain an Ably token embedded in a JWT token. Optional.
 *  - encrypted:   use '1' to obtain an encrypted token that embeds an Ably token. Optional.
 *  - returnType:  use 'jwt' to obtain a 'application/jwt' response, empty to obtain 'text/plain'. Optional.
 *  - environment: the environment in which Ably's requestToken is called. Either 'staging' or 'sandbox'.  Defaults to sandbox. Optional.
 *  - capability:  capability to associate to a token. Optional.

```

Examples/PublisherExample/Sources/Screens/Map/MapViewController.swift
QuintinWillison: My point is that code in a comment will never pass a compiler in CI. It is, therefore, at risk of falling out of sync with interface changes elsewhere and is a maintainability risk for the codebase. We should avoid this.

ben-xD: Oh I see what you mean, and in fact, I initially had an enum and switch 😂. However, it just complicated the code so much, and I removed it from the example app. A user would never have an enum or switch to choose between authentication methods. Then I would suggest just removing these lines from the example app.

Tests/SystemTests/Utils/JWTHelper.swift
lukasz-szyszkowski: if it's possible please resolve this conversation, this change will be done in scope of this issue https://github.com/ably/ably-asset-tracking-swift/issues/185#issuecomment-923656235

Examples/PublisherExample/Sources/Screens/Map/MapViewController.swift
ben-xD: I've removed the commented code and added documentation to the README instead 96109b031b5d285dfb316c71933af2286f47d922

QuintinWillison (approved):
I notice you've committed `DEVELOPMENT_TEAM` entries to Xcode projects. Was that intentional?

lukasz-szyszkowski (approved):

lukasz-szyszkowski:
> I notice you've committed `DEVELOPMENT_TEAM` entries to Xcode projects. Was that intentional?

no, it wasn't. I could remove it if it shouldn't be there

lukasz-szyszkowski:
> I notice you've committed `DEVELOPMENT_TEAM` entries to Xcode projects. Was that intentional?

@QuintinWillison removed here 7944c925495a18579f255004f53dc08be1e18d8d

PR #175, opened 2021-08-06
Add `.ruby-version`, to add support for users using rbenv and RVM

ben-xD:
I've added `.ruby-version` alongside `.tool-versions`, to add support for users using **rbenv** (a more common tool) abd **rvm** while still maintaining support asdf.


QuintinWillison (changes requested):
This doesn't work for me. If I start out on the `main` branch:

```
ably-asset-tracking-swift % ruby --version
ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-darwin20]

ably-asset-tracking-swift % git checkout feature/use-ruby-versions-instead-of-asdf-specific-file
Branch 'feature/use-ruby-versions-instead-of-asdf-specific-file' set up to track remote branch 'feature/use-ruby-versions-instead-of-asdf-specific-file' from 'origin'.
Switched to a new branch 'feature/use-ruby-versions-instead-of-asdf-specific-file'

ably-asset-tracking-swift % ruby --version
No version set for command ruby
Consider adding one of the following versions in your config file at /Users/quintinwillison/.tool-versions
ruby 2.7.2
ruby 2.7.0
ruby 3.0.0

?126 ably-asset-tracking-swift % git checkout main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.

ably-asset-tracking-swift % ruby --version
ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-darwin20]

ably-asset-tracking-swift % asdf --version
v0.8.1-a1ef92a
```

I specifically don't have a `global` Ruby configured on my workstation.

ben-xD:
Oh, according to [Migrating from another Ruby version manager](https://github.com/asdf-vm/asdf-ruby#migrating-from-another-ruby-version-manager),
> To ease migration, you can have it read an existing .ruby-version file to find out what version of Ruby should be used. To do this, add the following to $HOME/.asdfrc:
> ```
> legacy_version_file = yes
> ```

I personally think rbenv would be a better fit since it is more common (and asdf users can add that to their rc file). My analysis is https://github.com/ably/ably-flutter/pull/137#discussion_r659543004

lukasz-szyszkowski:
@ben-xD @QuintinWillison this is quite old PR, are we waiting for something with this one?

ben-xD:
Nothing is waiting on me @lukasz-szyszkowski. I still suggest we use **rbenv**, since it is more commonly used, and asdf users can enable it if they really want to.

QuintinWillison:
My preference is to conform to using `.tool-versions` across our client libraries. It is used elsewhere internally too, for example by @SimonWoolf with realtime codebases.

ben-xD:
Okay, in that case we can close this.

IMHO, it is better to use a more popular and easier tool designed specifically for Ruby (rbenv), instead of using the lowest common denominator manager for all possible tools (e.g. using asdf for ruby, node, etc.). As you can tell, I don't like asdf 😅. 

Generally though, I'm afraid that aiming to **standardise** (e.g. readmes, tools, github workflow file naming conventions) between repos before tradeoffs are understood can be detrimental and restrict the improvement/ growth than each repo/ project can make, compared to if they didn't have this standardisation cost. For example, calling everything `check.yml` is not very helpful, and each workflow should have a meaningful name.

QuintinWillison:
I am open to having _both_ files present, perhaps adding a note to the contributing guide (oh, ok, even that's not here yet - https://github.com/ably/ably-asset-tracking-swift/issues/171) on upgrading Ruby and the need to indicate that in multiple places.

Out of interest, @ben-xD, why don't you like ASDF?

There are multiple people across Ably org who are using it successfully, including myself. It allows us to conform the way we manage our tools in an unsurprising manner from repository context to repository context. What's not to like with that?

ben-xD:
Cool, now, in this PR's latest commit, we **have both files**.

---

Some opinions/ explanations: 

`rbenv` is a simpler and more popular tool and would prefer to use it myself, if I had to manage ruby. `asdf` tries to do too much, and doesn't do anything very well/ comfortably. I would prefer to keep `asdf` off my machine. 

The primary reason is asdf is _complicated_. Look at their getting started page: https://asdf-vm.com/guide/getting-started.html and them using plugins for every runtime you want to use. In a video call, you mentioned reading the help page of `asdf`, and all the information is there.

Both rbenv and [Ruby Version Manager (RVM)](https://rvm.io/) use `.ruby_version`. 

[rbenv](https://github.com/rbenv/rbenv#using-package-managers)? `brew install rbenv` then `rbenv init`. Isn't that simple?

SimonWoolf:
> The primary reason is asdf is complicated

There's different ways something can be complicated.

I'm not familiar with the internals of rbenv, but I did used to use rvm back when I used to be a rails dev. It works by replacing the `cd` shell builtin with its own shell function, which set up the right ruby environment in a shell when you changed into that directory. It's simple to set up, sure. But it's complicated in other, less obvious ways. Replacing `cd` might well work in most cases but will have weird failure modes; it violates the "[what if two programs did this?](https://devblogs.microsoft.com/oldnewthing/20050607-00/?p=35413)" rule. That's not just an aesthetic concern: if you're someone who's switching between lots of languages, each of which has its own separate version manager, some of which use mechanisms that only work if it's the only thing on the computer that does that, you now need to figure out exactly how each of them works to make sure they're not going to break each other.

asdf isn't as out-of-the-box-easy-to-use as some of the others, sure - the setup is more explicit, less automagic. It's uninvasive: it doesn't try and override or replace shell functionality or do anything magic on directory change, it just makes everything a symlink to a script which decides on and runs to the right version of ruby etc. at call time. It means you don't need to worry about different version managers clashing. And it means you only have to learn one version manager CLI instead of a dozen.

(Not trying to persuade you to switch -- there's no problem with supporting both -- just explaining the reasoning for my preference 🙂)

QuintinWillison (approved):

PR #181, opened 2021-08-27
Retry strategy

lukasz-szyszkowski:
It's the implementation of a retry strategy for failed requests.
I've added 2 more events for the event queue:
- `SendEnhancedLocationSuccessEvent` - enqueue event on send enhanced location success
- `SendEnhancedLocationFailureEvent` - enqueue event on send enhanced location failure
it's based on android implementation


KacperKluka (changes requested):
One important thing is missing - blocking sending of a message for a trackable if the previous sending hasn't finished. On Android, we have the `LocationsPublishingState` which holds the information about pending messages and we use it to postpone sending of any other messages for a given trackable :wink: 
This may not look like a problem now, but after introducing `skippedLocations` it will became one :sweat_smile:

lukasz-szyszkowski:
> One important thing is missing - blocking sending of a message for a trackable if the previous sending hasn't finished. On Android, we have the `LocationsPublishingState` which holds the information about pending messages and we use it to postpone sending of any other messages for a given trackable 😉
> This may not look like a problem now, but after introducing `skippedLocations` it will became one 😅

@KacperKluka I'll manage it in https://github.com/ably/ably-asset-tracking-swift/pull/182 (Send skipped locations PR)

KacperKluka (approved):
The pending messages behaviour will be implemented in https://github.com/ably/ably-asset-tracking-swift/pull/182 so this PR looks good :+1:

Sources/AblyAssetTrackingPublisher/PublisherTrackableState.swift
```diff
@@ -0,0 +1,44 @@
+ //
```
ben-xD, line 1: Just like Quintin mentioned (maybe in an unrelated PR), I think we should remove the Xcode autogenerated comments at the top, since they are not very useful. I remember at one place I worked, a developer mentioned that it was nice/ gave them pride to have their name at the top. I hope removing this autogenerated comment doesn't negatively affect morale 😅

Sources/AblyAssetTrackingPublisher/PublisherTrackableState.swift
```diff
@@ -0,0 +1,44 @@
...
+//  Created by Łukasz Szyszkowski on 25/08/2021.
+//
+
+import Foundation
+
+struct PublisherTrackableState {
+    
+    private enum Constants {
```
ben-xD, line 9: Why is this an enum? And im not sure, but would it make sense to make it customizable?

Sources/AblyAssetTrackingPublisher/PublisherTrackableState.swift
```diff
@@ -0,0 +1,44 @@
+ //
+//  Created by Łukasz Szyszkowski on 25/08/2021.
+//
+
+import Foundation
+
+struct PublisherTrackableState {
```
ben-xD, line 7: I have a programming question: When should you use classes vs mutable structs? aka. Why did you use `struct` instead of class?

Sources/AblyAssetTrackingPublisher/PublisherTrackableState.swift
lukasz-szyszkowski: - enum because you can't initialize it (you don't have to create `private init`)
- this is an internal class so `maxRetryCount` could be helpful for the future use and additionally I'm using it in tests

Sources/AblyAssetTrackingPublisher/PublisherTrackableState.swift
lukasz-szyszkowski: `struct` is OK here, it is lightweight and it's perfect for data "storage". It's a value type and their performance is better than `class` which is a reference type. 
Value types (`structs`) are stored on the `Stack` but the reference types (`class`) are stored on `Heap`, `Stack` is faster than `Heap`
The "problem" with structs is that they are [copied on write](https://medium.com/@nitingeorge_39047/copy-on-write-in-swift-b44949436e4f) - as I mentioned above, it's ok for use as a state.
more [HERE](https://abhimuralidharan.medium.com/difference-between-value-type-and-a-reference-type-in-ios-swift-18cb5145ad7a)
I know there is a `class` type state in my code, the only reason why I used them is for testing purposes. I injected them to the state manager and I can check their actual state (eg collection capacity) by reference. This is something that is hard to check when you are using `struct` which is copied on write. To be consistent I have to change all `class` types to `struct` types and change tests a little bit.

Tests/PublisherTests/DefaultPublisher/DefaultPublisherTests.swift
```diff
@@ -586,4 +593,87 @@ class DefaultPublisherTests: XCTestCase {
...
+        publisher.track(trackable: trackable) { _ in }
+        
+        wait(for: [trackCompletionHandlerExpectation], timeout: 5.0)
+        
+        let expectationDidUpdateLocation = self.expectation(description: "Publisher did update location expectation")
+        delegate.publisherDidUpdateEnhancedLocationCallback = { expectationDidUpdateLocation.fulfill() }
+        
+        /**
+         On every `sendEnhancedAssetLocation` request, the `sendEnhancedAssetLocationUpdateCounter` is incremented by `1`
+         Initialy we need to reset this counter
```
ben-xD, line 658: > Initialy we need to reset this counter

Why is that? Perhaps that would be good to add to comments, thanks :)

ben-xD (approved):
LGTM with comments, thanks 🙂

Sources/AblyAssetTrackingPublisher/PublisherTrackableState.swift
ben-xD: I meant why not just create a `private let maxRetryCount  = 1` inside `PublisherTrackableState`

Sources/AblyAssetTrackingPublisher/PublisherTrackableState.swift
ben-xD: All methods are mutating the struct, causing a new struct to be created. I don't think its fair to call it lightweight, or better performance at this point.

Sources/AblyAssetTrackingPublisher/PublisherTrackableState.swift
lukasz-szyszkowski: On second thought, I agree with you.
I used `Constants` "namespace" to separate this from other variables in this class, this is how we do it in my company, but, as we can see, this is unnecessary. I'll change it.

lukasz-szyszkowski: 3f490050915edcc38c1acb1410e3c68562b80e60

Sources/AblyAssetTrackingPublisher/PublisherTrackableState.swift
lukasz-szyszkowski: 8ff3ce15045fea1dcc47643f5e3c093af7984976

ben-xD: I guess this wasn't committed, I can't find this commit: https://github.com/ably/ably-asset-tracking-swift/pull/181/commits

lukasz-szyszkowski: 8ff3ce15045fea1dcc47643f5e3c093af7984976

lukasz-szyszkowski: if it's ok to you, please resolve conversation and we can merge this one

PR #182, opened 2021-08-31
Send skipped locations

lukasz-szyszkowski:


Sources/AblyAssetTrackingPublisher/State/DefaultSkippedLocationsState.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  Created by Łukasz Szyszkowski on 30/08/2021.
+//
+
+import Foundation
+
+class DefaultSkippedLocatoinsState: PublisherSkippedLocationsState {
```
KacperKluka, line 7: Typo - there is `Locatoins` but should be `Locations` :wink:

Sources/AblyAssetTrackingPublisher/State/DefaultSkippedLocationsState.swift
```diff
@@ -0,0 +1,36 @@
...
+    init(maxSkippedLocationsSize: Int = 60) {
+        self.maxSkippedLocationsSize = maxSkippedLocationsSize
+    }
+    
+    func add(trackableId: String, location: EnhancedLocationUpdate) {
+        var locations = skippedLocations[trackableId] ?? []
+        locations.append(location)
+        if locations.count > maxSkippedLocationsSize {
+            locations.remove(at: .zero)
+        }
```
KacperKluka, line 21: Shouldn't we sort locations by time to be sure that we're removing the oldest location update? :thinking:

Tests/PublisherTests/DefaultPublisher/DefaultPublisherTests.swift
```diff
@@ -628,52 +628,91 @@ class DefaultPublisherTests: XCTestCase {
         
     }
     
-    func testPublisherWillRetryOnFailureOnSendEnhancedLocationUpdate() {
+    func testDefaultSkippedLocationsState() {
```
KacperKluka, line 631: This test is quite long and has multiple asserts which leads me to think that it could be divided into more small tests :wink:

Tests/PublisherTests/Mocks/MockAblyPublisherService.swift
```diff
@@ -38,12 +38,15 @@ class MockAblyPublisherService: AblyPublisherService {
...
     var sendEnhancedAssetLocationUpdateParamTrackable: Trackable?
     var sendEnhancedAssetLocationUpdateParamCompletion: ResultHandler<Void>?
+    var sendEnhancedAssetLocationUpdateParamCompletionHandler: ((ResultHandler<Void>?) -> Void)?
     func sendEnhancedAssetLocationUpdate(locationUpdate: EnhancedLocationUpdate, forTrackable trackable: Trackable, completion: ResultHandler<Void>?) {
         sendEnhancedAssetLocationUpdateCounter += 1
         sendEnhancedAssetLocationUpdateCalled = true
         sendEnhancedAssetLocationUpdateParamLocationUpdate = locationUpdate
         sendEnhancedAssetLocationUpdateParamTrackable = trackable
         sendEnhancedAssetLocationUpdateParamCompletion = completion
+
```
KacperKluka, line 48: It looks like an unnecessary newline, maybe let's remove it? :wink:

KacperKluka (commented):
Overall looks good but I have a few questions :wink:

Sources/AblyAssetTrackingPublisher/State/DefaultSkippedLocationsState.swift
lukasz-szyszkowski: 1ad265d7795c5c5ea3680aa014b45b54cf160589

Sources/AblyAssetTrackingPublisher/State/DefaultSkippedLocationsState.swift
lukasz-szyszkowski: d174b97ea43d7666a97de45d02ca18a38f0aadaa

Tests/PublisherTests/DefaultPublisher/DefaultPublisherTests.swift
lukasz-szyszkowski: 2e7c30a2d6e2aaad90c38ced5a0842537e0aa4dd

Tests/PublisherTests/Mocks/MockAblyPublisherService.swift
lukasz-szyszkowski: 2e7c30a2d6e2aaad90c38ced5a0842537e0aa4dd

KacperKluka (approved):
LGTM :+1:

KacperKluka (changes requested):
I'm withdrawing my approval due to this https://github.com/ably/ably-asset-tracking-swift/pull/181#issuecomment-912298755
> One important thing is missing - blocking sending of a message for a trackable if the previous sending hasn't finished. On Android, we have the LocationsPublishingState which holds the information about pending messages and we use it to postpone sending of any other messages for a given trackable :wink:
This may not look like a problem now, but after introducing skippedLocations it will became one :sweat_smile:

KacperKluka (approved):
Looks good now :+1: One more thing perhaps that you could add is a test that checks if the waiting events are processed in the correct order (FIFO in our case) :wink:

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
```diff
@@ -65,7 +66,10 @@ class DefaultPublisher: Publisher {
          ablyService: AblyPublisherService,
          locationService: LocationService,
          routeProvider: RouteProvider,
-         trackableState: PublisherTrackableState = .init()) {
+         trackableState: PublisherTrackableState = DefaultTrackableState(),
+         skippedLocationsState: PublisherSkippedLocationsState = DefaultSkippedLocationsState()
```
ben-xD, line 70: Isn't the trackable updates stored in `PublisherSkippedLocationsState` fit nicer as part of `PublisherTrackableState`? These are locations which are for the trackable. It is rightfully its state. Ably-java does it the same way as you though.

Tests/PublisherTests/Mocks/MockResolutionPolicyFactory.swift
```diff
@@ -4,7 +4,9 @@ import AblyAssetTrackingCore
 class MockResolutionPolicyFactory: ResolutionPolicyFactory {
     var resolutionPolicy: MockResolutionPolicy?
     func createResolutionPolicy(hooks: ResolutionPolicyHooks, methods: ResolutionPolicyMethods) -> ResolutionPolicy {
-        resolutionPolicy =  MockResolutionPolicy(hooks: hooks, methods: methods)
+        if resolutionPolicy == nil {
```
ben-xD, line 7: - So `createResolutionPolicy` doesn't always create a `resolutionPolicy`? That's confusing/ bad name
- Why do we re-use it, is it really expensive? Or even worse, are we sharing state between different calls of `createResolutionPolicy`? I think we should avoid this singleton pattern if possible

ben-xD (commented):
I've given some comments which can be worked on in the mean time because I have not yet reviewed 4 files (tests related). I also want to follow through the logic in an IDE soon, since this is a domain I haven't worked with much.

Sources/AblyAssetTrackingPublisher/State/DefaultTrackableState.swift
```diff
@@ -0,0 +1,90 @@
+//
```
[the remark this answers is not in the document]
lukasz-szyszkowski: one word: testability :)
It can be mocked with ease in tests.

Protocol-oriented-programing is more flexible than OOP, below few benefits from [HERE](https://tsh.io/blog/protocol-oriented-programming-swift/):
- Types can conform to more than one protocol.
- They can also obtain default behaviours from multiple protocols.
- Unlike multiple inheritance in other programming languages, protocol extensions do not introduce any additional states.
- Protocols can be adopted by classes, structs and enums, while base classes and inheritance are restricted to class types only.
- Protocols allow retroactive modeling with extensions added to existing types.

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
lukasz-szyszkowski: All of this code is based on AAT Android (kotlin).
```kotlin
private inner class State(
        routingProfile: RoutingProfile,
        locationEngineResolution: Resolution
) {
    private var isDisposed: Boolean = false
    var isStopped: Boolean = false
    var locationEngineResolution: Resolution = locationEngineResolution
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    var isTracking: Boolean = false
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    val trackables: MutableSet<Trackable> = mutableSetOf()
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    val trackableStates: MutableMap<String, TrackableState> = mutableMapOf()
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    val trackableStateFlows: MutableMap<String, MutableStateFlow<TrackableState>> = mutableMapOf()
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    val lastChannelConnectionStateChanges: MutableMap<String, ConnectionStateChange> = mutableMapOf()
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    var lastConnectionStateChange: ConnectionStateChange = ConnectionStateChange(
        ConnectionState.OFFLINE, null
    )
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    val resolutions: MutableMap<String, Resolution> = mutableMapOf()
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    val lastSentEnhancedLocations: MutableMap<String, Location> = mutableMapOf()
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    val skippedEnhancedLocations: SkippedLocations = SkippedLocations()
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    var estimatedArrivalTimeInMilliseconds: Long? = null
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    var lastPublisherLocation: Location? = null
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    var currentDestination: Destination? = null
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    val subscribers: MutableMap<String, MutableSet<Subscriber>> = mutableMapOf()
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    val requests: MutableMap<String, MutableMap<Subscriber, Resolution>> = mutableMapOf()
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    var presenceData: PresenceData = PresenceData(ClientTypes.PUBLISHER)
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    var active: Trackable? = null
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
        set(value) {
            this@DefaultCorePublisher.active = value
            field = value
        }
    var routingProfile: RoutingProfile = routingProfile
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
        set(value) {
            this@DefaultCorePublisher.routingProfile = value
            field = value
        }
    val rawLocationChangedCommands: MutableList<(State) -> Unit> = mutableListOf()
        get() = if (isDisposed) throw PublisherStateDisposedException() else field
    val enhancedLocationsPublishingState: LocationsPublishingState = LocationsPublishingState()
        get() = if (isDisposed) throw PublisherStateDisposedException() else field

    fun dispose() {
        trackables.clear()
        trackableStates.clear()
        trackableStateFlows.clear()
        lastChannelConnectionStateChanges.clear()
        resolutions.clear()
        lastSentEnhancedLocations.clear()
        skippedEnhancedLocations.clearAll()
        estimatedArrivalTimeInMilliseconds = null
        active = null
        lastPublisherLocation = null
        currentDestination = null
        subscribers.clear()
        requests.clear()
        rawLocationChangedCommands.clear()
        enhancedLocationsPublishingState.clearAll()
        isDisposed = true
    }
}
```

Sources/AblyAssetTrackingPublisher/State/DefaultTrackableState.swift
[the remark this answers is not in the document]
ben-xD: I am aware of protocol oriented programming. That still means the protocol should be designed well, my question above details design issues regardless of OOP or POP, and I don't think you've addressed them. 

Also, there is only one implementation of `PublisherTrackableState` anyway (I cannot find a mock implementation of it). If you did implement a mock, it would have to implement redundant implementation details (increment, reset counter etc) which would be code duplication.

Also, you should avoid mocks where possible. `DefaultTrackableState` is a very simple implementation, why mock it?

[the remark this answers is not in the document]
ben-xD: I understand protocol oriented programming. As we both know, that still means the protocol should be designed well, my question above details design issues regardless of OOP or POP, and I don't think you've addressed them. (`DefaultX` naming, inheritance over protocols in this case). Its been a few months since I watched [the video](https://developers.apple.com/videos/play/wwdc2015/408/), but I've watched a few times over the last few years.

---

Now I am taking a look at `TestableTrackableState.swift`, and I confused. There are **setter** functions called `removeWillCall` and `removeDidCall`. They're not being used at all? IMHO It's a very unreadable class with 6 optional closures. I don't know what you're testing here.

It seems to me you are implementing a mock just so you can check the method was called. You pass a closure to this mock which will be called back when the method gets called, e.g.
```
        trackableState.unmarkMessageAsPendingDidCall {
            unmarkMessageAsPendingDidCallExpectation.fulfill()
        }
```

You are already inheriting `DefaultTrackableState` with `class TestableTrackableState: DefaultTrackableState`, so the protocol is not actually useful. Inheriting makes sense, because you don't want code duplication. I don't think adding this protocol is adding anything except more lines of code at the moment. Why not have `PublisherTrackableState` the original implementation, and `TestablePublisherTrackableState` inherit this, without protocols?

The fact a `XState` class is a protocol is already bad isn't it? It doesn't sound like a protocol. 

Also, SOLID principle: Interface Segregation Principle (ISP) is broken. This is quite a large interface (11 methods).

---

## Potential solution

Anyway, to solve this testing problem, lets think of the public API of `PublisherTrackableState`. Would anyone reasonably want to know what the state is? It is a `state` after all. We could make it less of an opaque box and allow users to check if something is in it or not? Then in your tests, you can just call this method/ field and check it. It sounds like a useful method even outside of testing.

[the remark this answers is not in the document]
ben-xD: In general though, IMHO there is a problem with the API of this entity, regardless of its a protocol or a class.

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
ben-xD: I already guessed you based it off AATA:
>  I think Ably-java does it the same way as you though.

I'm not sure if you agree or disagree with my first comment in this thread 😅?

Sources/AblyAssetTrackingPublisher/State/DefaultTrackableState.swift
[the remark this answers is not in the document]
lukasz-szyszkowski: I've made a bit of changes here d3e9b72fe26d04ee46cb2f2af7a6b0065417a147
- I've split protocol into few smaller protocols (ISP problem)
- I've used multiple protocol inheritances
- I've removed unnecessary code from `TestableTrackableState` and added a comment

WDYT?

Tests/PublisherTests/Mocks/MockResolutionPolicyFactory.swift
lukasz-szyszkowski: 182e59433c22a24e407b597f08e716a30bc22648

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
lukasz-szyszkowski: 96b0db747bbf5f6938a0a3a18533c1a14b035154

Tests/PublisherTests/Mocks/MockResolutionPolicyFactory.swift
ben-xD: We don't need a field though, do we? It can just be returned from the function immediately? In tests, just save a reference to `resolutionPolicy` returned by `createResolutionPolicy`. I think that would be a nice simplification.

Sources/AblyAssetTrackingPublisher/State/DefaultTrackableState.swift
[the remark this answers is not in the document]
ben-xD: It's mostly looking simpler I think. I think it's mostly good, but I think ISP is not the main problem with that interface. Would anyone realistically use those protocols separately (`typealias TrackableStateable = StateWaitable & StatePendable & StateRemovable & StateRetryable & StateSkippable`)? The main issue is that it leaks implementation details about how a class which implements it would work. IMHO That is the problem. This protocol split makes the code even more unreadable. 😢

---

Okay, tbh I did not **suggest** anything. I only asked a question:
> Isn't the trackable updates stored in PublisherSkippedLocationsState **fit nicer** as part of PublisherTrackableState? These are locations which are for the trackable. It is rightfully its state.

Let me elaborate. :) Don't assume I want you to do anything, because it should be your decision at the end of the day. I was asking what you thought, would it fit better? Don't do it if you disagree, or explain your thoughts too. It should be for discussion/ thoughts, I'm not telling you to do anything. I'm not sure if you agree or disagree with my thoughts quoted above.

Tests/PublisherTests/Mocks/MockResolutionPolicyFactory.swift
lukasz-szyszkowski: this is not my idea, and since I'm not 100% sure that this change will not affect the test, I don't want to change it. This should be done here https://github.com/ably/ably-asset-tracking-swift/issues/185#issuecomment-923656235

ben-xD (approved):

PR #183, opened 2021-09-02
Send notification about trip started on a meta channel

lukasz-szyszkowski:


KacperKluka (commented):
I don't think that this should be merged as we've decided that we won't send that data from the SDKs :sweat_smile: Look here https://github.com/ably/ably-asset-tracking-android/issues/431 @AndyNicks could you please confirm if this PR should not be merged? :bow:

PR #191, opened 2021-09-28
Add missing `completion?(.success)` closure call

lukasz-szyszkowski:


QuintinWillison (approved):
It's great that you spotted this, but I'll admit I'm intrigued why:

1. there wasn't a test in place that would have already revealed this oversight
2. you've not added a test now to validate your fix

lukasz-szyszkowski:
> It's great that you spotted this, but I'll admit I'm intrigued why:
> 
> 1. there wasn't a test in place that would have already revealed this oversight
> 2. you've not added a test now to validate your fix

All of the tests are running on mocked objects, there was no possibility to detect this bug until the `System Tests` were introduced [HERE](https://github.com/ably/ably-asset-tracking-swift/pull/169/files).
There is a test called `PublisherAndSubscriberSystemTests` which uses real-life Ably Service. This test sometimes worked sometimes did not, I thought the test has been not very well written and I decided to fix it [HERE](https://github.com/ably/ably-asset-tracking-swift/pull/186), but after deeper code investigation I've found a bug that I mentioned in this PR. 
So, the fix is here - in this PR and the test already exists in `main` and is enhanced [HERE](https://github.com/ably/ably-asset-tracking-swift/pull/186)

ben-xD (approved):

PR #193, opened 2021-10-05
Update Mapbox dependency to public release

lukasz-szyszkowski:


KacperKluka (approved):

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
```diff
@@ -17,6 +17,7 @@ class DefaultLocationService: LocationService {
         }
 
         let directions = Directions(credentials: mapboxConfiguration.getCredentians())
+        NavigationSettings.shared.initialize(directions: directions, tileStoreConfiguration: .default)
```
ben-xD, line 20: Why is this added here, in a PR to bump dependency version? It's a bit confusing to see this here since this line seems like it either should be present already to make it work, or it would be unnecessary.

Package.swift
```diff
@@ -19,11 +19,11 @@ let package = Package(
...
     dependencies: [
-        .package(name: "MapboxNavigation", url: "https://github.com/mapbox/mapbox-navigation-ios.git", .exact("2.0.0-beta.22")),
+        .package(name: "MapboxNavigation", url: "https://github.com/mapbox/mapbox-navigation-ios.git", .exact("2.0.0-rc.4")),
         .package(url: "https://github.com/realm/SwiftLint", from: "0.43.1"),
         .package(url: "https://github.com/apple/swift-log", from: "1.4.2"),
         .package(url: "https://github.com/ably/ably-cocoa", from: "1.2.6")
-        ],
+    ],
     targets: [
         //
```
ben-xD, line 28: If you remove `test targest`, why not remove this `// Targets` too.

ben-xD (changes requested):

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
lukasz-szyszkowski: it's a Mapbox migration change - it needs to be here

ben-xD: > it needs to be here
What happens without it?

lukasz-szyszkowski: an exception, where Mapbox requires an entry in Info.plist

ben-xD (approved):
Approved, if the comment is removed (conversation above) to be consistent/clean in `Package.swift`

QuintinWillison:
I've invited @lawrence-forooghian to this repository as I would like to get him exposed to this codebase also. Once he's accepted that invite then I would like him to review this pull request.

Tests/SystemTests/PublisherAuthenticationSystemTests.swift
```diff
@@ -121,7 +121,7 @@ class PublisherAuthenticationSystemTests: XCTestCase {
             .connection(configuration)
             .mapboxConfiguration(MapboxConfiguration(mapboxKey: Secrets.mapboxAccessToken))
             .log(logConfiguration)
-            .locationSource(LocationSource(locationSource: [CLLocation(latitude: 0.0, longitude: 0.0)]))
+            .locationSource(LocationSource(locationSource: [CLLocation(latitude: 0.0, longitude: 0.0), CLLocation(latitude: 1.0, longitude: 1.0)]))
```
lawrence-forooghian, line 124: Do we want to revert this once Mapbox release version 2.1 of their SDK? If so, we should create an issue to track this.

lawrence-forooghian (commented):
From my limited understanding of the codebase, this seems sensible to me. Just one question.

Tests/SystemTests/PublisherAuthenticationSystemTests.swift
lukasz-szyszkowski: No, we won't revert it, this is totally fine for testing purposes.

lawrence-forooghian (approved):

lukasz-szyszkowski:
This change e0e9e1c5d59ef21e0b139969bce4b56660a70c04 is in the case of the Mapbox token configuration via `NavigationSettings.shared.initialize` where documentation says:
> If you want to use this method, it should be the first method you use from Navigation SDK.
> Not doing so will lead to undefined behavior

.github/workflows/check.yml
```diff
@@ -25,3 +25,10 @@ jobs:
         env:
           MAPBOX_ACCESS_TOKEN: ${{ secrets.MAPBOX_ACCESS_TOKEN }}
           ABLY_API_KEY: ${{ secrets.ABLY_API_KEY }}
+
+      - name: Xcodebuild Logs Artifact
+        if: always()
+        uses: actions/upload-artifact@v2
+        with:
+          name: xcodebuild-logs
+          path: ~/Library/Developer/Xcode/DerivedData/*/Logs
```
KacperKluka, line 34: If I understand your [comment](https://github.com/ably/ably-asset-tracking-swift/pull/193#issuecomment-951634212) correctly, this code was added only for some investigation reasons, right? Shouldn't we remove it then?

KacperKluka (approved):
LGTM :+1: I just have one small questions :wink:

PR #199, opened 2021-10-27
Move .xcworkspace to `Examples` directory

lukasz-szyszkowski:
### INFO:
_All of the checks of this PR will be failed because moving the `.xcworkspace` makes the Fastlane targets inaccessible.
There is another PR (https://github.com/ably/ably-asset-tracking-swift/pull/200) on top of this commit that removes `Fastlane` from the project towards the raw `xcodebuild` command wrapped in the shell script._

PR #200, opened 2021-10-28
Remove fastlane

lukasz-szyszkowski:
Linked PR https://github.com/ably/ably-asset-tracking-swift/pull/199


.github/workflows/assemble.yml
```diff
@@ -30,7 +30,5 @@ jobs:
 
       - name: Build All Projects
         run: |
-          bundle install
-          fastlane build_subscriber
-          fastlane build_publisher
-          fastlane build_example_apps
+            bundle install
+            ./Scripts/./assemble.sh
```
KacperKluka, line 34: I think you could remove the `./` in the middle :wink: So it would look like this `./Scripts/assemble.sh`

.github/workflows/check.yml
```diff
@@ -18,10 +18,8 @@ jobs:
         env:
           MAPBOX_DOWNLOADS_TOKEN: ${{ secrets.MAPBOX_DOWNLOADS_TOKEN }}
 
-      - name: Run All Tests using Fastlane
-        run: |
-          bundle install
-          fastlane test_all
+      - name: Run All Tests
+        run: ./Scripts/./test.sh
```
KacperKluka, line 22: Like above, I think you could remove the `./` in the middle and it would look like this `./Scripts/test.sh` :wink:

KacperKluka (approved):
Looks good, just two suggestions which does not change the way in which the changes work :wink:

ben-xD (approved):

.github/workflows/assemble.yml
lukasz-szyszkowski: here 7605cbe9eed6324568ac98188ee2faa9ea2c2c70

.github/workflows/check.yml
lukasz-szyszkowski: here 7605cbe9eed6324568ac98188ee2faa9ea2c2c70

PR #204, opened 2021-11-03
Fix .ruby-version

lawrence-forooghian:
I think this file was just copied from .tool-versions, but rbenv doesn’t know what to do with the "ruby" bit.


QuintinWillison (approved):

lukasz-szyszkowski (approved):

lukasz-szyszkowski:
@lawrence-forooghian, the `check.yml` won't pass because the Mapbox navigation lib beta version produces errors. This problem does not arise in the latest (2.0.0) version

ben-xD (approved):

PR #209, opened 2021-11-19
Add ConnectionConfiguration convenience initialiser visible in obj-c

lukasz-szyszkowski:
Since `public convenience init(clientId: String? = nil, authCallback: @escaping AuthCallback)` is not visible in objective-c because of using swift `Result<T,U>` I had to create a separated initialiser for objective-c. This "small" change required further changes in the codebase.


Sources/AblyAssetTrackingCore/Configuration/ConnectionConfiguration.swift
```diff
@@ -10,46 +10,92 @@ public enum AuthResult {
...
+    public let objcAuthCallback: ObjCAuthCallback?
     
     /**
      Connect to Ably using basic authentication (API Key)
      
      - Parameters:
-       - apiKey: API key string obtained from application dashboard.
-       - clientId: Optional identifier to be assigned to this client.
-       - authCallback: A callback that will be used to authenticate with Ably, including at initial connection and for renewing an expired token.
+     - apiKey: API key string obtained from application dashboard.
```
ben-xD, line 52: I noticed this indentation change, and this made me think what formatting are we using? DocC/Swift has guidance for this at https://www.swift.org/documentation/docc/writing-symbol-documentation-in-your-source-files, you might be interested :) 

Resolve this conversation whenever you want 😇

lukasz-szyszkowski: I use the `conrtol+i` shortcut and the indent is set to 4 spaces / 4 tabs

lukasz-szyszkowski: do you think `///` is better than `/** ... */` ?
Is `///` recommended by Apple (or swift community) now?

Sources/AblyAssetTrackingInternal/Configuration/ConnectionConfiguration+Extensions.swift
```diff
@@ -46,8 +49,34 @@ extension ConnectionConfiguration {
                 }
             })
         }
-
+        
         return authCallbackWrapper
     }
-
+    
+    private func createObjCAuthCallbackWrapper(_ authCallback: @escaping ObjCAuthCallback) -> (ARTTokenParams, @escaping (ARTTokenDetailsCompatible?, NSError?) -> ()?) -> () {
```
ben-xD, line 56: Should it be `createAuthCallback(from: objCAuthCallback)`? Just a suggestions :)

AblyAssetTracking.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -172,6 +172,15 @@
           "version": "3.1.1"
         }
       },
+      {
+        "package": "swift-argument-parser",
```
ben-xD, line 176: I don't see any uses of `import ArgumentParser`, I'm wondering if this dependency is needed or accidental?

KacperKluka (approved):

Sources/AblyAssetTrackingInternal/Configuration/ConnectionConfiguration+Extensions.swift
lukasz-szyszkowski: I don't like this naming too, but we have to be sure what this method is for, and this one is for Objective-C

Sources/AblyAssetTrackingCore/Configuration/ConnectionConfiguration.swift
ben-xD: In dart, the linter tells me to use `///`. I think it is a little bit better because each line can be moved around independently. However, DocC says both is supported and **doesn't recommend anything specific**.

However, shouldn't the parameters be children of `- Parameters`?. Maybe Xcode is not good at indenting docs 😅. If we follow the DocC guidelines, we have a better chance of generating DocC as well in the future.

Sources/AblyAssetTrackingCore/Configuration/ConnectionConfiguration.swift
```diff
@@ -10,46 +10,92 @@ public enum AuthResult {
     case tokenDetails(TokenDetails)
 }
 
+public class ObjcAuthResultJWT: NSObject {
```
ben-xD, line 13: All 3 classes below are just wrappers for the underlying field they hold. This requires users to wrap their result in `authCallback` explicitly. However, why not just use String, TokenRequest and TokenDetails directly? They are all unique types.

ben-xD (approved):

Sources/AblyAssetTrackingInternal/Configuration/ConnectionConfiguration+Extensions.swift
ben-xD: But the method parameter being `@escaping ObjCAuthCallback` ensures this already. 

My **thought** when I first read the method name was not what the method does: It is creating wrapper for AuthCallback so it can be used in Objective-C (wrong). Instead of creating an AuthCallback wrapper for an Objective-C AuthCallback (the actual method implementation). But, then I thought, the signature already explains this behaviour already.

just a minor point

lukasz-szyszkowski: of course, I could name it the same since they have different argument types - I forgot it's swift 🙉
I'll rename them to `createAuthCallback`

Thanks

AblyAssetTracking.xcworkspace/xcshareddata/swiftpm/Package.resolved
lukasz-szyszkowski: it was added before my joining this project, and to be honest, I don't know what is it for. I'll investigate it and will remove it if it's unnecessary. But it should be done in separate PR

An Issue: https://github.com/ably/ably-asset-tracking-swift/issues/211

Sources/AblyAssetTrackingInternal/Configuration/ConnectionConfiguration+Extensions.swift
lukasz-szyszkowski: commit f6249cfefb0be006573966890e1154e564333068

Sources/AblyAssetTrackingCore/Configuration/ConnectionConfiguration.swift
lukasz-szyszkowski: here c98f8a223d3216a189341cb6b06191f73a65f8a5

Sources/AblyAssetTrackingCore/Configuration/ConnectionConfiguration.swift
lukasz-szyszkowski: here 94caaf50cd1ed4c483d577adf9887374f61d25cb

ben-xD: THanks 🚀

Sources/AblyAssetTrackingInternal/Configuration/ConnectionConfiguration+Extensions.swift
ben-xD: Cool

Sources/AblyAssetTrackingCore/Configuration/ConnectionConfiguration.swift
ben-xD: Cool!

PR #219, opened 2021-12-07
Remove Objective-C related code

lukasz-szyszkowski:
In few classes I should remove `NSObject` inheritance.
I left it because some of them requires functionality of `NSObject` by default, e.g.: Hashable, Equatable, debugDescription and description

I'll remove it in separate PR


Package.resolved
```diff
@@ -42,8 +42,8 @@
         "repositoryURL": "https://github.com/mapbox/mapbox-directions-swift.git",
         "state": {
           "branch": null,
-          "revision": "f95be4f764acb0ea40ab11184e09825d3046d14f",
```
lawrence-forooghian, line 45: Do these dependency changes need to be in this commit?

Sources/AblyAssetTrackingCore/Configuration/ConnectionConfiguration.swift
```diff
@@ -62,15 +50,6 @@ public class ConnectionConfiguration: NSObject {
...
-    @objc
-    public convenience init(clientId: String? = nil, objcAuthCallback: @escaping ObjCAuthCallback) {
-        self.init(
-            apiKey: nil,
-            clientId: clientId,
-            objcAuthCallback: objcAuthCallback
-        )
-    }
-    
     @objc
```
lawrence-forooghian, line 53: Can we remove this `@objc` too?

Examples/AblyAssetTracking.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -14,9 +14,9 @@
         "package": "Amplify",
         "repositoryURL": "https://github.com/aws-amplify/amplify-ios",
         "state": {
-          "branch": "main",
```
lawrence-forooghian, line 17: Do these dependency changes need to be in this commit?

lawrence-forooghian (commented):
A couple of other observations:

- There seems to be a problem with the `assemble` GitHub workflow – you haven't removed the Objective-C examples from `Scripts/assemble.sh` but the workflow is still passing. Looking at the workflow logs, there are definitely things failing inside that script. I think you'll need to investigate why workflow isn't hence failing.
- I think you can also remove the `@objc` on `MapboxConfiguration.init(mapboxKey:)`.

lawrence-forooghian:
Ah, and I think it would also be good to remove references to Objective-C from `README.md` and `CONTRIBUTING.md`.

Examples/AblyAssetTracking.xcworkspace/xcshareddata/swiftpm/Package.resolved
lukasz-szyszkowski: it's automatically managed by Xcode when I'm opening an Example app. 
I probably shouldn't push it but I'll leave it here.

Sources/AblyAssetTrackingCore/Configuration/ConnectionConfiguration.swift
lukasz-szyszkowski: b40d1fa2472a89845d37775e13630befcd0a4118

Package.resolved
lukasz-szyszkowski: > A couple of other observations:
> 
> * There seems to be a problem with the `assemble` GitHub workflow – you haven't removed the Objective-C examples from `Scripts/assemble.sh` but the workflow is still passing. Looking at the workflow logs, there are definitely things failing inside that script. I think you'll need to investigate why workflow isn't hence failing.
> * I think you can also remove the `@objc` on `MapboxConfiguration.init(mapboxKey:)`.

b40d1fa2472a89845d37775e13630befcd0a4118

ben-xD (approved):

lukasz-szyszkowski:
> Ah, and I think it would also be good to remove references to Objective-C from `README.md` and `CONTRIBUTING.md`.

here 81583a1da7061e2a0daa64fb581cf5c08cbe030d

maratal (approved):

lawrence-forooghian:
> There seems to be a problem with the assemble GitHub workflow – you haven't removed the Objective-C examples from Scripts/assemble.sh but the workflow is still passing. Looking at the workflow logs, there are definitely things failing inside that script. I think you'll need to investigate why workflow isn't hence failing.

@lukasz-szyszkowski from what I can tell, you've removed the Objective-C build steps, but not made any changes to the rest of the build process. But were you able to investigate why the failure of these build steps didn't cause the CI job to fail? Because it seems like there might be a broader problem. If not something you want to fix here, I think it definitely needs an issue.

lukasz-szyszkowski:
> > There seems to be a problem with the assemble GitHub workflow – you haven't removed the Objective-C examples from Scripts/assemble.sh but the workflow is still passing. Looking at the workflow logs, there are definitely things failing inside that script. I think you'll need to investigate why workflow isn't hence failing.
> 
> @lukasz-szyszkowski from what I can tell, you've removed the Objective-C build steps, but not made any changes to the rest of the build process. But were you able to investigate why the failure of these build steps didn't cause the CI job to fail? Because it seems like there might be a broader problem. If not something you want to fix here, I think it definitely needs an issue.

To be honest, I don't know what you mean, there are two commits:
https://github.com/ably/ably-asset-tracking-swift/pull/219/commits/b40d1fa2472a89845d37775e13630befcd0a4118 remove building objc examples
https://github.com/ably/ably-asset-tracking-swift/commit/81583a1da7061e2a0daa64fb581cf5c08cbe030d remove references to objc in README and CONTRIBUTING

lawrence-forooghian:
> To be honest, I don't know what you mean

So, my original comment was this:

> There seems to be a problem with the `assemble` GitHub workflow – you haven't removed the Objective-C examples from `Scripts/assemble.sh` but the workflow is still passing. Looking at the workflow logs, there are definitely things failing inside that script. I think you'll need to investigate why workflow isn't hence failing.

So what I'm saying is that when I wrote that comment, the following three things were all true:

- the Objective-C examples no longer existed;
- the `assemble` GitHub workflow still contained commands to try and build the Objective-C examples;
- the CI build was green.

I was saying that we need to understand why the CI build was passing even though the Objective-C examples no longer existed. It suggests that the script might not be correctly checking for the failure of its build steps.

lukasz-szyszkowski:
> > To be honest, I don't know what you mean
> 
> So, my original comment was this:
> 
> > There seems to be a problem with the `assemble` GitHub workflow – you haven't removed the Objective-C examples from `Scripts/assemble.sh` but the workflow is still passing. Looking at the workflow logs, there are definitely things failing inside that script. I think you'll need to investigate why workflow isn't hence failing.
> 
> So what I'm saying is that when I wrote that comment, the following three things were all true:
> 
> * the Objective-C examples no longer existed;
> * the `assemble` GitHub workflow still contained commands to try and build the Objective-C examples;
> * the CI build was green.
> 
> I was saying that we needed to understand why the CI build was passing even though the Objective-C examples no longer existed. It suggests that the script might not be correctly checking for the failure of its build steps.

Good catch! Thanks.

PR for this https://github.com/ably/ably-asset-tracking-swift/pull/226
POC here https://github.com/ably/ably-asset-tracking-swift/pull/224

CONTRIBUTING.md
```diff
@@ -76,9 +76,9 @@ The release process must include the following steps:
...
 ## Coding Conventions and Style Guide
 
-- The SDKs are written in Swift, however they still have to be compatible for use from Objective-C based apps.
+- The SDKs are written in Swift.
 - Favor Protocol Oriented Programming with Dependency Injection when writing any code. We're unable to create automatic mocks in Swift, so it'll be helpful for writing unit tests.
 - SwiftLint is integrated into the project. Make sure that your code does not add any SwiftLint related warning.
 - Please remove default Xcode header comments (with author, license and creation date) as they're not necessary.
 - If you're adding or modifying any part of the public interface of SDK, please also update [QuickHelp](https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_markup_formatting_ref/SymbolDocumentation.html#//apple_ref/doc/uid/TP40016497-CH51-SW1) documentation.
-- Docs are written for both Swift and ObjC
+- Docs are written for Swift
```
lawrence-forooghian, line 84: I'm not sure that the two modified bullet points make sense after removing the mention of Objective-C. Do you think they can just be removed now?

lukasz-szyszkowski: d0c5da66930d285e2d181270b8ae5ee56fff7475

lukasz-szyszkowski:
[Why does it fail?](https://github.com/ably/ably-asset-tracking-swift/pull/229)

CONTRIBUTING.md
lawrence-forooghian: I was also referring to "The SDKs are written in Swift."

lukasz-szyszkowski: > I was also referring to "The SDKs are written in Swift."

be80427ea18338b69ed1a1a348b773dd9de4f352

lawrence-forooghian (approved):

PR #220, opened 2021-12-07
Remove `NSObject` inheritance

lukasz-szyszkowski:


Sources/AblyAssetTrackingCore/Model/ErrorInformation.swift
```diff
@@ -122,3 +122,20 @@ public class ErrorInformation: NSObject, Error, CustomNSError {
         self.href = nil
     }
 }
+
+extension ErrorInformation: Hashable {
```
lawrence-forooghian, line 126: Do the types modified in this PR need to be classes? They seem like good candidates for structs to me. And then we can use the default implementations of `Hashable`, `Equatable`.

lukasz-szyszkowski: I changed it to the struct, but we can't use the default implementation of `Equatable` and `Hashable` because
>It only works for structs or enums (not classes) and then only when each stored property of a struct or associated value of an enum is itself Equatable or Hashable.

and the `public let cause: Error?` isn't `Equatable/Hashable`

changed here 57b89b9e66ff99a2b259aa8c7c8ebe7ce7f78a24

lawrence-forooghian: Sorry if I wasn't clear, but I meant changing classes to structs for all the classes touched by this PR, not just `ErrorInformation`.

As for `Equatable`, `Hashable` etc – is it just `ErrorInformation` where you can't use the default implementation? Could we use it for the other types in this PR?

And I wonder whether it even makes sense for `ErrorInformation` to conform to `Hashable`, if `Error` doesn't. What is the use case?

lukasz-szyszkowski: It's used in tests, e.g.:
https://github.com/ably/ably-asset-tracking-swift/blob/3d09ee69d9d31fd08c96b3d379cfebd94bd2f006/Tests/PublisherTests/DefaultPublisher/DefaultPublisherTests.swift#L185
https://github.com/ably/ably-asset-tracking-swift/blob/3d09ee69d9d31fd08c96b3d379cfebd94bd2f006/Tests/PublisherTests/DefaultPublisher/DefaultPublisherTests.swift#L485
>Sorry if I wasn't clear, but I meant changing classes to structs for all the classes touched by this PR, not just ErrorInformation.

I would like to do this in separate PR, WDYT?

lawrence-forooghian: > I would like to do this in separate PR, WDYT?

The reason I suggested doing it here is that it avoids us needing to add all of the `Equatable` / `Hashable` implementations, only to then remove them again in the next PR.

lawrence-forooghian: > It's used in tests, e.g.

I think it's probably more important to consider whether it makes sense for a user of the library. I can't think of an occasion in which a user would want to compare two `ErrorInformation` values. I think that if it's just used by the tests, better to just have a helper function in the test file to do the comparison.

ben-xD (approved):

maratal (approved):

Sources/AblyAssetTrackingCore/Model/ErrorInformation.swift
lawrence-forooghian: @lukasz-szyszkowski did you have any thoughts on this?

lukasz-szyszkowski: > @lukasz-szyszkowski did you have any thoughts on this?

yea, I'll remove Hashable/Equatable from the ErrorInformation struct and I'll create some helper for testing purposes.

lukasz-szyszkowski: @lawrence-forooghian here a476609341dbb72b495098bc2127b4ac3bcddb58

Tests/PublisherTests/DefaultPublisher/DefaultPublisherTests.swift
```diff
@@ -861,3 +861,12 @@ class DefaultPublisherTests: XCTestCase {
         waitForExpectations(timeout: 10.0)
     }
 }
+
+extension ErrorInformation {
```
lawrence-forooghian, line 865:
```suggestion
private extension ErrorInformation {
```

lawrence-forooghian: Or, if you're using it in more than one test file, perhaps put it in a test helpers file?

lawrence-forooghian: Also, `isEqual` is a misleading name I think, since it doesn't compare the `cause`. Perhaps a name that makes it clear what you want to check is the same about the two errors?

lukasz-szyszkowski: > Also, `isEqual` is a misleading name I think, since it doesn't compare the `cause`. Perhaps a name that makes it clear what you want to check is the same about the two errors?

what's your proposal then?

lawrence-forooghian: `isEqualIgnoringCause(to:)`?

lukasz-szyszkowski: I've reverted this commit in case of https://github.com/ably/ably-asset-tracking-swift/runs/4534264133?check_suite_focus=true#step:4:3861

lawrence-forooghian: Sorry, I don't understand why you reverted it? Isn't the solution 

> Or, if you're using it in more than one test file, perhaps put it in a test helpers file?

lukasz-szyszkowski: Sorry for the confusion, I've moved it to separated file 25da7e28529756d238551dd2be3a44052317475d
About method naming, I want to leave it as it is `isEqual` is fine IMO.

lawrence-forooghian (approved):

PR #229, opened 2021-12-16
Add `xcode-select` to github actions step

lukasz-szyszkowski:


lukasz-szyszkowski:
> Can we consider long term issues, where we are stuck with an older version of Xcode once an ever newer version of Xcode is released?

We can remove `xcode-select` when 13.2 will be a default version. WDYT?

maratal:
Shouldn't it be in a separate step? How to check that this version was installed?

lukasz-szyszkowski:
> Shouldn't it be in a separate step? How to check that this version was installed?

Good point, I've change it here 7173e4f1fec0498d54f69068fff48e924fe473b0

maratal (approved):

lukasz-szyszkowski:
Same problem in another PR https://github.com/ably/ably-asset-tracking-swift/runs/4578965682?check_suite_focus=true#step:5:9541

lawrence-forooghian (approved):

PR #233, opened 2021-12-27
SwiftUI Publisher Example App

lukasz-szyszkowski:
This app is not a 1:1 UIKit Publisher App. The UI is simplified (for example users don't need to tap a button to request location authorization, it can be done automatically)

App features:
- request user location using CoreLocation framework
- start and stop `Ably Asset Tracking Swift Publisher`
- track `trackable` with the destination (this could be changed within the code) and without it
- changing the routing profile to `driving`, `walking` etc

https://user-images.githubusercontent.com/47628502/148733772-f5210009-d409-4daf-8057-58e33b8d04cb.mp4


lawrence-forooghian (commented):
Hey @lukasz-szyszkowski, I haven't looked at the code for this yet. It's quite a big one so it would be useful to have some guidance on how to review this. For example:

- what is the functionality that it offers – could you for example provide a video?
- is it meant to be exactly the same as the non-SwiftUI example app? If not, how does it differ?

Also I was wondering whether we're happy to take on the additional work of maintaining an extra example app? Do you think that we need to keep both UIKit and SwiftUI examples?

Are we planning to add a subscriber example app too?

lukasz-szyszkowski:
> Also I was wondering whether we're happy to take on the additional work of maintaining an extra example app? Do you think that we need to keep both UIKit and SwiftUI examples?

My answer is yes. It's good to have example apps for currently used iOS technology where SwiftUI is a more futuristic approach than UIKit. Maintaining it is not a big deal, and we are sure that `ably-cococa` and `ably aat swift` are working properly with SwiftUI.

> Are we planning to add a subscriber example app too?

I don't think we need to create a SwiftUI example since we are sure that everything works properly just having the Publisher App. But if you think it's necessary (and why it is?) I can create some simple SwiftUI Subscriber App

Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/Style/CustomTextFieldStyle.swift
```diff
@@ -0,0 +1,28 @@
...
+import SwiftUI
+
+struct CustomTextFieldStyle: TextFieldStyle {
+    private let isEnabled: Bool
+    
+    init(isEnabled: Bool = true) {
+        self.isEnabled = isEnabled
+    }
+    
+    func _body(configuration: TextField<Self._Label>) -> some View {
```
maratal, line 12: Don't know much about SwiftUI, but why is it `_body` here and `makeBody` in `CustomButtonStyle`? Just curious.

lukasz-szyszkowski: > Don't know much about SwiftUI, but why is it `_body` here and `makeBody` in `CustomButtonStyle`? Just curious.

To be honest, me neither :) but here is a bit of explanation of underscored properties in SwiftUI https://stackoverflow.com/a/65209375 don't know yet too much about SwiftUI but I guess this is a good point in time to learn more about it :D

maratal: Swift nowadays is a hell of a learning ride! 🤪

maratal (approved):

Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/Style/CustomTextFieldStyle.swift
lukasz-szyszkowski: true story bro! :D

PR #244, opened 2022-01-12
Create custom `Location` struct 

lukasz-szyszkowski:
The idea behind these changes is to create custom `Location` object to be independent of `CoreLocation` objects


Sources/AblyAssetTrackingCore/Model/Location.swift
```diff
@@ -0,0 +1,100 @@
+
+import Foundation
+
+public struct Location: Equatable {
+    
+    public let coordinate: LocationCoordinate
```
KacperKluka, line 6: Every other field has a comment but this one doesn't, please add a comment for this field :wink:

KacperKluka (changes requested):
Just one small thing :wink:

Sources/AblyAssetTrackingCore/Model/Location.swift
lukasz-szyszkowski: > Every other field has a comment but this one doesn't, please add a comment for this field 😉

added here e6d96ed5ae10cb04c958b27897912a546674995a

KacperKluka (approved):
LGTM :+1:

maratal (approved):

PR #248, opened 2022-01-25
Handle resolutions for the location engine in the Publisher

lukasz-szyszkowski:


Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
```diff
@@ -33,7 +33,7 @@ class DefaultLocationService: LocationService {
     }
 
     func changeLocationEngineResolution(resolution: Resolution) {
-        // TODO: Implement method
+        locationManager.systemLocationManager.desiredAccuracy = resolution.accuracy.toCoreLocationAccuracy()
```
KacperKluka, line 36: Awesome that you don't have to rewrite the whole location manager to do that :muscle: However, two things are missing (desired interval and minimum displacement), look at how it's implemented in the [Android code](https://github.com/ably/ably-asset-tracking-android/blob/main/publishing-sdk/src/main/java/com/ably/tracking/publisher/locationengine/MappingFunctions.kt#L7-L11) :wink:

KacperKluka (changes requested):
Looks good but it's missing something :wink:

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
lukasz-szyszkowski: Ok, I forgot about `distanceFilter` I'll add it soon, but there is no way to specify something like `time intervals` between checking locations in CLLocationManager

lukasz-szyszkowski: Update here 9422f28c264115162a959e39403be54f722a2e5d

KacperKluka: It looks better now, thanks :bow: It's a shame that we can't modify the time interval :disappointed: I think it's worth mentioning somewhere in the docs or API as this might affect some of users expectations, WDYT @AndyNicks ? :thinking:

AndyNicks: Yes, it is definitely has to be mentioned in the documentation. I think it has to me bentioned in README as well

lukasz-szyszkowski: I'm not sure if I need to mention it in README.md @AndyNicks, actually, there is no mention of `Resolution` so I would leave it as it is. Who I should inform in case of documentation update?

KacperKluka: AFAIK there's already an issue created for this problem, therefore, I'm resolving this conversation :wink:

KacperKluka (approved):
LGTM :+1:

AndyNicks (approved):

Sources/AblyAssetTrackingCore/Extension/Accuracy+CoreLocation.swift
```diff
@@ -0,0 +1,22 @@
...
+        case .minimum:
+            if #available(iOS 14, *) {
+                return kCLLocationAccuracyReduced
+            } else {
+                return kCLLocationAccuracyThreeKilometers
+            }
+        case .low:
+            return kCLLocationAccuracyKilometer
+        case .balanced:
+            return kCLLocationAccuracyNearestTenMeters
```
maratal, line 15: I think the amount of accuracy steps should match the one Apple gives us, so probably worth adding one more, f.e. `.sufficient` for kCLLocationAccuracyNearestTenMeters, and reduce `.balanced` to hundreds meters.

lukasz-szyszkowski: This is a good idea but these enums are "untouchable" because it's designed some time ago to work on iOS, Android and many other platforms for AAT, so if I want to add new enum, every other platform should do the same moreover the documentation should be updated.
This PR is responsible just for mapping AAT Lib values to CoreLocation equivalents. If you don't feel alright with `kCLLocationAccuracyNearestTenMeters` for `.balanced` I would change it but please keep in mind there is no clear explanation why balanced should be mapped to `ten meters` or `hundred meters`  Or maybe there is? :)

maratal (approved):

PR #266, opened 2022-02-23
Send raw locations from the Publisher

### LINKED ISSUE #250: Send raw locations from the Publisher
Enabled sending of the raw locations on the same channel on which the enhanced locations are being sent. The raw locations have all the mechanisms that the enhanced ones have, especially resolution and skippedLocations. As the raw locations shouldn't be normally published, there is a new option on the builder that enables the raw locations. The commentary on this method states clearly that this should not be enabled in the production environment. Additionally, information about whether the raw locations are enabled is added to the PresenceData of the publisher. see https://github.com/ably/ably-asset-tracking-android/pull/493

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-1533) by [Unito](https://www.unito.io)

lukasz-szyszkowski:
closes #250

### DIFF
--- a/Sources/AblyAssetTrackingInternal/PresenceData.swift
+++ b/Sources/AblyAssetTrackingInternal/PresenceData.swift
@@ -4,14 +4,17 @@ import AblyAssetTrackingCore
 public struct PresenceData: Codable {
     public let type: ClientType
     public let resolution: Resolution?
+    public let rawLocations: Bool?
 
-    public init(type: ClientType) {
+    public init(type: ClientType, rawLocations: Bool? = nil) {
         self.type = type
         self.resolution = nil
+        self.rawLocations = rawLocations
     }
 
-    public init(type: ClientType, resolution: Resolution?) {
+    public init(type: ClientType, resolution: Resolution?, rawLocations: Bool? = nil) {
         self.type = type
         self.resolution = resolution
+        self.rawLocations = rawLocations
     }
 }
--- a/Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
+++ b/Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
@@ -577,7 +577,7 @@ extension DefaultPublisher {
     }
     
     private func performRawLocationChanged(_ event: RawLocationChangedEvent) {
-        guard !state.isStoppingOrStopped else {
+        guard !state.isStoppingOrStopped, areRawLocationsEnabled == true else {
             logger.error("Cannot perform EnhancedLocationChangedEvent. Publisher is not working.")
             return
         }
@@ -602,10 +602,8 @@ extension DefaultPublisher {
             return shouldSend
         }
 
-        if areRawLocationsEnabled == true {
-            trackablesToSend.forEach { trackable in
-                sendRawLocationUpdate(event: event, trackable: trackable)
-            }
+        trackablesToSend.forEach { trackable in
+            sendRawLocationUpdate(event: event, trackable: trackable)
         }
     }
     
--- a/Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
+++ b/Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
@@ -32,7 +32,7 @@ class DefaultPublisher: Publisher {
     private let batteryLevelProvider: BatteryLevelProvider
     
     private var ablyPublisher: AblyPublisher
-    private var trackableState: TrackableState<EnhancedLocationUpdate>
+    private var enhancedLocationState: TrackableState<EnhancedLocationUpdate>
     private var state: State = .working
     private var presenceData = PresenceData(type: .publisher)
 
@@ -67,7 +67,7 @@ class DefaultPublisher: Publisher {
          ablyPublisher: AblyPublisher,
          locationService: LocationService,
          routeProvider: RouteProvider,
-         trackableState: TrackableState<EnhancedLocationUpdate> = TrackableState()
+         enhancedLocationState: TrackableState<EnhancedLocationUpdate> = TrackableState()
     ) {
         
         self.connectionConfiguration = connectionConfiguration
@@ -78,7 +78,7 @@ class DefaultPublisher: Publisher {
         self.locationService = locationService
         self.ablyPublisher = ablyPublisher
         self.routeProvider = routeProvider
-        self.trackableState = trackableState
+        self.enhancedLocationState = enhancedLocationState
 
         self.batteryLevelProvider = DefaultBatteryLevelProvider()
 
@@ -341,7 +341,7 @@ extension DefaultPublisher {
             
             switch result {
             case .success(let wasPresent):
-                self?.trackableState.remove(trackableId: event.trackable.id)
+                self?.enhancedLocationState.remove(trackableId: event.trackable.id)
                 wasPresent
                     ? self?.enqueue(event: ClearRemovedTrackableMetadataEvent(trackable: event.trackable, resultHandler: event.resultHandler))
                     : self?.callback(value: false, handler: event.resultHandler)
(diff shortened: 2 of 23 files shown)


Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
```diff
@@ -527,21 +550,98 @@ extension DefaultPublisher {
...
+            }
+            
+            return shouldSend
+        }
+
+        if areRawLocationsEnabled == true {
+            trackablesToSend.forEach { trackable in
+                sendRawLocationUpdate(event: event, trackable: trackable)
+            }
+        }
```
KacperKluka, line 609: Here you are performing some logic to filter the trackables and check if a raw location update for them should be sent even if the raw locations are disabled :wink: I'd say that the "are raw locations enabled" check should be done as soon as possible, so we won't waste performance on things we don't need :thinking:

Tests/InternalTests/PresenceDataTests.swift
```diff
@@ -5,15 +5,33 @@ import UIKit
 
 class PresenceDataTests: XCTestCase {
     func testSerializationPublisher() throws {
-        let data = PresenceData(type: .publisher)
-        let jsonString = try data.toJSONString()
+        var data = PresenceData(type: .publisher, rawLocations: true)
+        var jsonString = try data.toJSONString()
+        XCTAssertNotNil(jsonString)
```
KacperKluka, line 10: Why are you only checking if the `jsonString` is not null? I think we should be able to check if it has been parsed to a string in a correct format :thinking:

KacperKluka (changes requested):
Looks good, just a few small things :wink:

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
lukasz-szyszkowski: 5d8105d928c5c22aacee7c83b57e3c98089c098e

Tests/InternalTests/PresenceDataTests.swift
lukasz-szyszkowski: 5ac7ec651a4ee2c6d8b00e11fc4e66ed6abe5bc9

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
KacperKluka: It's better but now you would report that the publisher has failed to send a message because it is not working, while in the reality it can't send it because it's simply disabled :wink: So I'd move it to a separate guard with its own error log message.

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
```diff
@@ -527,21 +550,96 @@ extension DefaultPublisher {
...
+        enhancedLocationState.addLocation(for: trackableId, location: location)
+    }
+    
+    private func saveRawLocationForFurtherSending(trackableId: String, location: RawLocationUpdate) {
+        rawLocationState.addLocation(for: trackableId, location: location)
+    }
+    
+    private func performRawLocationChanged(_ event: RawLocationChangedEvent) {
+        guard !state.isStoppingOrStopped, areRawLocationsEnabled == true else {
+            logger.error("Cannot perform EnhancedLocationChangedEvent. Publisher is not working.")
```
KacperKluka, line 581: The error message seems to be wrong :wink: 
```suggestion
            logger.error("Cannot perform RawLocationChangedEvent. Publisher is not working.")
```

KacperKluka (changes requested):
Almost there :grin:

Tests/PublisherTests/DefaultPublisher/DefaultPublisher_LocationServiceTests.swift
```diff
@@ -199,7 +199,7 @@ class DefaultPublisher_LocationServiceTests: XCTestCase {
             publisher: publisher,
             locationUpdate: locationUpdate,
             trackable: trackable,
-            trackableState: trackableState,
+                    enhancedLocationState: trackableState,
```
maratal: looks like extra indentation

Sources/AblyAssetTrackingInternal/PresenceData.swift
```diff
@@ -4,14 +4,17 @@ import AblyAssetTrackingCore
 public struct PresenceData: Codable {
     public let type: ClientType
     public let resolution: Resolution?
+    public let rawLocations: Bool?
 
-    public init(type: ClientType) {
+    public init(type: ClientType, rawLocations: Bool? = nil) {
```
maratal, line 9: name `rawLocations` implies an array of locations. I'd suggest `isRawLocations`

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
```diff
@@ -67,7 +71,9 @@ class DefaultPublisher: Publisher {
          ablyPublisher: AblyPublisher,
          locationService: LocationService,
          routeProvider: RouteProvider,
-         enhancedLocationState: TrackableState<EnhancedLocationUpdate> = TrackableState()
+         enhancedLocationState: TrackableState<EnhancedLocationUpdate> = TrackableState(),
+         rawLocationState: TrackableState<RawLocationUpdate> = TrackableState(),
+         areRawLocationsEnabled: Bool? = nil
```
maratal, line 76: Isn't it just yes/no? Why optional used?

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
```diff
@@ -48,6 +48,7 @@ extension DefaultLocationService: PassiveLocationManagerDelegate {
     }
     
     func passiveLocationManager(_ manager: PassiveLocationManager, didUpdateLocation location: CLLocation, rawLocation: CLLocation) {
+        delegate?.locationService(sender: self, didUpdateRawLocationUpdate: RawLocationUpdate(location: rawLocation.toLocation()))
```
maratal, line 51: What if raw locations are turned off?

Sources/AblyAssetTrackingInternal/PresenceData.swift
lukasz-szyszkowski: This is imposed by JSON schema. Of course, I can map it on encoding but is this worth changing?

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
lukasz-szyszkowski: taken from android implementation, don't know if "third" value (the `nil`) will be used in future? @KacperKluka ?

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
lukasz-szyszkowski: delegation works no matter if `raw locations` are enabled or not

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
KacperKluka: I don't remember exactly but it was probably connected with backwards compatibility :thinking:

lukasz-szyszkowski: Is @KacperKluka response is enough to close this conversation?

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
lukasz-szyszkowski: aa699882c9338a466ad8ab6a1f8555a15b5796b6

KacperKluka (approved):
Looks good now, thanks :+1:

Tests/PublisherTests/DefaultPublisher/DefaultPublisher_LocationServiceTests.swift
maratal: what about this one?

maratal (approved):

PR #268, opened 2022-02-28
Replace `CLLocation` with AAT Wrapper `Location` in subscriber delegate

lukasz-szyszkowski:
According to weird commits history in this PR. I'm closing it.
New PR with this change and proper history can be find here  #271

### DIFF
--- a/Sources/AblyAssetTrackingCore/Extension/GeoJSONMessage+Location.swift
+++ b/Sources/AblyAssetTrackingCore/Extension/GeoJSONMessage+Location.swift
@@ -5,7 +5,7 @@ public extension GeoJSONMessage {
         return Location(
             coordinate: LocationCoordinate(latitude: geometry.latitude, longitude: geometry.longitude),
             altitude: geometry.altitude,
-            ellipsoidalAltitude: .zero,
+            ellipsoidalAltitude: -1,
             horizontalAccuracy: properties.accuracyHorizontal,
             verticalAccuracy: properties.accuracyVertical ?? -1,
             course: properties.bearing ?? -1,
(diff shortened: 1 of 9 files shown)


Sources/AblyAssetTrackingCore/Extension/GeoJSONMessage+Location.swift
```diff
@@ -0,0 +1,18 @@
...
+        return Location(
+            coordinate: LocationCoordinate(latitude: geometry.latitude, longitude: geometry.longitude),
+            altitude: geometry.altitude,
+            ellipsoidalAltitude: .zero,
+            horizontalAccuracy: properties.accuracyHorizontal,
+            verticalAccuracy: properties.accuracyVertical ?? -1,
+            course: properties.bearing ?? -1,
+            courseAccuracy: properties.accuracyBearing ?? -1,
+            speed: properties.speed ?? -1,
+            speedAccuracy: properties.accuracySpeed ?? -1,
```
KacperKluka, line 14: Are you sure that those `-1` won't break anything? [On Android](https://github.com/ably/ably-asset-tracking-android/blob/main/publishing-sdk/src/main/java/com/ably/tracking/publisher/LocationHistoryModels.kt#L20-L37) we're using `0` as the default values :wink:

lukasz-szyszkowski: to be honest I'm not 100% sure 😅  I'll change it for consistency

lukasz-szyszkowski: 269f1435c399760796aea6074d52b40324b8f5cd

PR #269, opened 2022-02-28
Add raw locations to the Subscriber SDK

lukasz-szyszkowski:


KacperKluka (approved):
Everything looks good but I'm wondering whether we should use the `CLLocation` for the raw ones or maybe due to this PR https://github.com/ably/ably-asset-tracking-swift/pull/268 we should change it to `Location`? :thinking: I don't have enough context so I'll let you decide :wink:

lukasz-szyszkowski:
> > Everything looks good but I'm wondering whether we should use the `CLLocation` for the raw ones or maybe due to this PR #268 we should change it to `Location`? 🤔 I don't have enough context so I'll let you decide 😉
> 
> you're right, we should avoid using `CLLocation` everywhere it is possible. I'll change it.

changed here

PR #273, opened 2022-03-09
Send resolution to subscriber

lukasz-szyszkowski:


Examples/SubscriberExample/Sources/Screens/Map/MapViewController.swift
```diff
@@ -224,4 +224,8 @@ extension MapViewController: SubscriberDelegate {
         assetStatusLabel.textColor = statusDescAndColor.color
         assetStatusLabel.text = statusDescAndColor.desc
     }
+    
+    func subscriber(sender: Subscriber, didUpdateEnhancedLocation resolution: Resolution) {
+        // TODO
+    }
```
KacperKluka, line 230: Why is it added here? :thinking: Also the second param name probably shouldn't be `didUpdateEnhancedLocation` :sweat_smile:

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
```diff
@@ -356,6 +370,36 @@ extension DefaultAbly: AblyPublisher {
         }
     }
     
+    public func sendResolution(
+        trackable: Trackable,
+        resolution: Resolution,
+        completion: ResultHandler<Void>?
+    ) {
+        guard let channel = channels[trackable.id] else {
```
KacperKluka, line 378: We should not send the resolution data via channel messages, we should be using channel's presence :sweat_smile: On Android, we also went with the channel messages approach, however, it generated a few issues and we've decided to switch to presence data (more details here https://github.com/ably/ably-asset-tracking-android/issues/638) :wink:

KacperKluka (changes requested):
We should be sending the resolution data via channel presence data and not as channel messages. :wink: The requirements have changed and it looks like the iOS issue was not updated, sorry about that :disappointed:

Sources/AblyAssetTrackingPublisher/DefaultPublisherBuilder.swift
```diff
@@ -22,7 +23,9 @@ class DefaultPublisherBuilder: PublisherBuilder {
                  routingProfile: RoutingProfile?,
                  delegate: PublisherDelegate?,
                  resolutionPolicyFactory: ResolutionPolicyFactory?,
-                 areRawLocationsEnabled: Bool?) {
+                 areRawLocationsEnabled: Bool?,
+                 isSendResolutionEnabled: Bool = false
```
KacperKluka, line 27: Like above, I think that sending resolution should be enabled by default :wink:

Tests/SystemTests/ChannelModesTests.swift
```diff
@@ -159,6 +159,8 @@ extension ChannelModesTests: SubscriberDelegate {
     
     func subscriber(sender: AblyAssetTrackingSubscriber.Subscriber, didChangeAssetConnectionStatus status: ConnectionState) {}
     
+    func subscriber(sender: AblyAssetTrackingSubscriber.Subscriber, didUpdateResolution resolution: Resolution) {}
+
```
KacperKluka, line 163: Since you made that method optional, can't we remove it? :wink:

KacperKluka (changes requested):
Almost there :grin:

Sources/AblyAssetTrackingPublisher/DefaultPublisherBuilder.swift
lukasz-szyszkowski: changed a288514997ea39ae67ca98329302cc66012e4595

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
```diff
@@ -73,7 +74,8 @@ class DefaultPublisher: Publisher {
          routeProvider: RouteProvider,
          enhancedLocationState: TrackableState<EnhancedLocationUpdate> = TrackableState(),
          rawLocationState: TrackableState<RawLocationUpdate> = TrackableState(),
-         areRawLocationsEnabled: Bool? = nil
+         areRawLocationsEnabled: Bool? = nil,
+         isSendResolutionEnabled: Bool = false
```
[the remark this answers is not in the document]
lukasz-szyszkowski: changed a288514997ea39ae67ca98329302cc66012e4595

Tests/SystemTests/ChannelModesTests.swift
lukasz-szyszkowski: changed a288514997ea39ae67ca98329302cc66012e4595

KacperKluka (approved):
LGTM :+1:

Sources/AblyAssetTrackingSubscriber/DefaultSubscriber.swift
```diff
@@ -297,4 +298,9 @@ extension DefaultSubscriber: AblySubscriberServiceDelegate {
         logger.debug("subscriberService.didReceiveEnhancedLocation.", source: String(describing: Self.self))
         callback(event: DelegateEnhancedLocationReceivedEvent(location: location))
     }
+    
+    func subscriberService(sender: AblySubscriber, didReceiveResolution resolution: Resolution) {
+        logger.debug("subscriberService.didReceiveResolution.", source: String(describing: Self.self))
```
maratal, line 303: Cosmetics, but why not "\(Self.self)"?

lukasz-szyszkowski: no reason :) if it's a blocker I can change it if not I would leave it as it is

maratal (approved):

PR #274, opened 2022-03-10
Add ability to observe raw locations change in `Subscriber` SDK

lukasz-szyszkowski:


Sources/AblyAssetTrackingSubscriber/SubscriberDelegate.swift
```diff
@@ -20,6 +20,15 @@ public protocol SubscriberDelegate: AnyObject {
         - location: Received location.
      */
     func subscriber(sender: Subscriber, didUpdateEnhancedLocation location: CLLocation)
+    
+    /**
+     Called when the `Subscriber` receive any Raw Location (matched to road) update for observed trackable
```
KacperKluka, line 25: I don't think that the raw locations are "matched to road" :wink: I'd remove it from this documentation
```suggestion
     Called when the `Subscriber` receive any Raw Location update for observed trackable
```

Examples/SubscriberExample/Sources/Screens/Map/MapViewController.swift
```diff
@@ -218,10 +218,15 @@ extension MapViewController: SubscriberDelegate {
         self.location = location
         scrollToReceivedLocation()
     }
+    
+    func subscriber(sender: Subscriber, didUpdateRawLocation location: Location) {
+        // TODO
+    }
```
KacperKluka, line 224: Why is it added here? :thinking: I'd remove it if it's not necessary :wink:

KacperKluka (changes requested):
I have a few small suggestions :wink:

Examples/SubscriberExample/Sources/Screens/Map/MapViewController.swift
lukasz-szyszkowski: as I mentioned in other PR where you ask the same, there is no `optional` methods in Swift protocols, but I can create an extension with default implementation of this methos. This will work as an `optional`.

changed here 768614547668ffcd78a5b0bca650072e60b93f19

KacperKluka (approved):
LGTM :+1:

maratal (approved):

PR #281, opened 2022-03-22
Add next location update intervals to the Subscriber API

lukasz-szyszkowski:


Sources/AblyAssetTrackingSubscriber/SubscriberDelegate.swift
```diff
@@ -38,6 +38,11 @@ public protocol SubscriberDelegate: AnyObject {
         - resolution: Received `Resolution` object.
      */
     func subscriber(sender: Subscriber, didUpdateResolution resolution: Resolution)
+    
+    /**
+     Called when the `Subscriber` receive estimated next location update intervals (in milliseconds) for observed trackable
+     */
+    func subscriber(sender: Subscriber, didUpdateDesiredInterval interval: Double)
```
KacperKluka, line 45: Since the desired interval is in milliseconds why are you using `Double` and not `Long` here? :thinking:

lukasz-szyszkowski: I've used the original type from `Resolution` 
https://github.com/ably/ably-asset-tracking-swift/blob/ecb544f7f74a8de2d19541b7b5bbc6ed31e81d29/Sources/AblyAssetTrackingCore/Model/Resolution.swift#L17-L24

KacperKluka (approved):
LGTM :+1:

AndyNicks (approved):

PR #282, opened 2022-03-23
Upgrade Mapbox dependencies

lukasz-szyszkowski:
Refactor deprecated Mapbox code

### DIFF
--- a/Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
+++ b/Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
@@ -901,7 +901,6 @@ extension DefaultPublisher: AblyPublisherServiceDelegate {
 // MARK: ResolutionPolicyMethodsDelegate
 extension DefaultPublisher: DefaultResolutionPolicyMethodsDelegate {
     func resolutionPolicyMethods(refreshWithSender sender: DefaultResolutionPolicyMethods) {
-        resolutionPolicy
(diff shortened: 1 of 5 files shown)


AndyNicks (approved):

maratal (approved):

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
```diff
@@ -901,7 +901,6 @@ extension DefaultPublisher: AblyPublisherServiceDelegate {
 // MARK: ResolutionPolicyMethodsDelegate
 extension DefaultPublisher: DefaultResolutionPolicyMethodsDelegate {
     func resolutionPolicyMethods(refreshWithSender sender: DefaultResolutionPolicyMethods) {
-        resolutionPolicy
```
lawrence-forooghian, line 904: I'm a bit confused by what this line has to do with this PR, but I'm also a bit confused by what this line was doing before?

lukasz-szyszkowski: Well 😅 I caught this up by accident, and to be honest, this is some orphan that did nothing here. I don't know why this wasn't caught before.

lawrence-forooghian: OK - probably would be better as a separate PR (or at least a separate commit) but I think we can live with it 🙂

PR #284, opened 2022-03-23
Make `raw locations enabled` optional

lukasz-szyszkowski:


Sources/AblyAssetTrackingPublisher/DefaultPublisherBuilder.swift
```diff
@@ -171,7 +171,7 @@ class DefaultPublisherBuilder: PublisherBuilder {
                                        constantLocationEngineResolution: constantLocationEngineResolution)
     }
     
-    func rawLocations(enabled: Bool) -> PublisherBuilder {
+    func rawLocations(enabled: Bool?) -> PublisherBuilder {
```
maratal, line 174: With [this comment](https://github.com/ably/ably-asset-tracking-swift/pull/277/files#r832964412) I'm not sure now if it should be optional. Maybe change declaration of the property itself?

lukasz-szyszkowski: That's funny because [here](https://github.com/ably/ably-asset-tracking-swift/pull/266#discussion_r814229688) is a slightly different explanation 😅 @KacperKluka would you look at those comments, please?

KacperKluka: Okay, so we have 2 places to consider here: the API and the implementation :wink: 
- In [the API](https://github.com/ably/ably-asset-tracking-android/blob/main/publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt#L215) of the builder the `rawLocations` method does not accept nullable values but is optional meaning that you don't have to call the method if you don't need to. 
- In [the implementation](https://github.com/ably/ably-asset-tracking-android/blob/main/publishing-sdk/src/main/java/com/ably/tracking/publisher/PublisherBuilder.kt#L23) of the builder the field is nullable and the default value is null. I can't remember why exactly it is `null` and not `false` by default, but my guess is that it had something to do with backwards compatibility :thinking: 

Does it make sense? :sweat_smile:

lawrence-forooghian: Yeah, seems like it should remain non-nullable here, and I think that we should change the [implementation of `DefaultPublisher`](https://github.com/ably/ably-asset-tracking-swift/blob/7da1d4ed710cbf04130a38a8c27d6cc30750bff5/Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift#L40) to make it non-nullable…

lukasz-szyszkowski: changed here 9166e2c21a9b2cd47995ea936f5653af2c9c212e it's non `Optional` anymore.

lawrence-forooghian (approved):

PR #291, opened 2022-04-12
Remove Swift version of `Publisher Example App`

lukasz-szyszkowski:


maratal:
The title is a bit misleading tbh. I think it should be "Remove `UIKit` version of Publisher Example App". I approve regardless.

maratal (approved):

PR #296, opened 2022-05-20
Display the horizontal accuracy of locations in the subscriber example app

lukasz-szyszkowski:
https://user-images.githubusercontent.com/47628502/169485014-533641ab-f971-46dd-bf41-0fa4a8849b7e.mov


Examples/SubscriberExample/Sources/Screens/Map/MapViewController.swift
```diff
@@ -136,6 +134,27 @@ class MapViewController: UIViewController {
...
+            
+            if let view = mapView.view(for: annotation) as? HorizontalAccuracyAnnotationView {
+                view.accuracy = Double(rect.size.width)
+            }
+            
+            annotation.coordinate = coordinate
+        } else {
+            let annotation = HorizontalAccuracyAnnotation()
+            annotation.coordinate = coordinate
+            annotation.accuracy = position.accuracy
```
KacperKluka, line 154: Both in the `if` and `else` you are setting the `annotation.coordinate` and `annotation.accuracy`, maybe it would be possible to extract it out of this `if/esle` to avoid code duplication? :thinking:  Something like: 
(Warning: pseudo-Swift below :sweat_smile: )
```
if let annotation = mapView.annotations.first(where: { $0 is HorizontalAccuracyAnnotation }) as? HorizontalAccuracyAnnotation  {
    if let view = mapView.view(for: annotation) as? HorizontalAccuracyAnnotationView {
        view.accuracy = Double(rect.size.width)
    }
    } else {
        let annotation = HorizontalAccuracyAnnotation()
        mapView.addAnnotation(annotation)
    }
annotation.coordinate = coordinate
annotation.accuracy = position.accuracy
```

Examples/SubscriberExample/Sources/Screens/Map/MapViewController.swift
```diff
@@ -224,16 +249,24 @@ extension MapViewController: MKMapViewDelegate {
         scheduleResolutionUpdate()
     }
     
-    private func createAnnotationView(for annotation: TruckAnnotation) -> MKAnnotationView {
-        let annotationView = getAnnotationView(for: annotation)
+    private func createTruckAnnotationView(for annotation: TruckAnnotation) -> TruckAnnotationView {
+        let annotationView: TruckAnnotationView = getAnnotationView(for: annotation)
         annotationView.bearing = annotation.bearing
-        annotationView.backgroundColor = UIColor.blue.withAlphaComponent(0.7)
+        annotationView.zPriority = .init(rawValue: 1.0)
```
KacperKluka, line 255: I wonder whether it would be worth extracting this magic number to a constant called `FOREGROUND_Z` and then do the same for the `.zero` below and call it `BACKGROUND_Z`, WDYT? :thinking:

KacperKluka (changes requested):
Looks good but I've suggested a few small improvements :wink:

Examples/SubscriberExample/Sources/Screens/Map/MapViewController.swift
lukasz-szyszkowski: I've extracted annotation creation to the generic function. Changed here c21e66965c43f6eb9a53d314a4253544a03db53c

Examples/SubscriberExample/Sources/Screens/Map/MapViewController.swift
lukasz-szyszkowski: Since we have only `foreground` and `background` views we can use predefined values `min` and `max`. Changed here 711666eb8cccea39fd8af90f2bd917db398acd3a

KacperKluka: The `min/max` approach is a change for the better but I still think that it would be even better to use the `foreground/backround` one :wink: The reason is that to understand the `min/max` you need to know if max priority means foreground and min means background or is it the other way around :sweat_smile: WDYT? :wink:

lukasz-szyszkowski: you're right, changed here 41ce76c6221a73b907ff59ebb61ffcb1ba344c6e

KacperKluka (approved):
LGTM :+1:

Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/Model/SimulatedLocations.swift
```diff
@@ -107,7 +107,7 @@ struct SimulatedLocations {
             return CLLocation(
                 coordinate: .init(latitude: Double(values.first!)!, longitude: Double(values.last!)!),
                 altitude: 100,
-                horizontalAccuracy: Double(2 + arc4random_uniform(20)), //accuracy between 2m and 22m
+                horizontalAccuracy: index < 10 ? 25 : 35,
```
maratal, line 110: What are 25 and 35? Maybe move these into constants?

Examples/SubscriberExample/Sources/Screens/Map/MapAnnotation/TruckAnnotationView.swift
```diff
@@ -1,7 +1,8 @@
 import Foundation
 import MapKit
 
-class TruckAnnotationView: MKAnnotationView {
+class TruckAnnotationView: MKAnnotationView, Identifiable {
```
maratal, line 4: Truck? Was it heavy vehicle meant?)

lukasz-szyszkowski: I've just added `Identifiable` conformance. This naming is not my idea.

Examples/SubscriberExample/SubscriberExample.xcodeproj/project.pbxproj
```diff
@@ -425,7 +445,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				ASSETCATALOG_COMPILER_GLOBAL_ACCENT_COLOR_NAME = AccentColor;
 				CODE_SIGN_STYLE = Automatic;
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = C4YGD8Z6G2;
```
maratal, line 448: Was it on purpose?

Examples/SubscriberExample/SubscriberExample.xcodeproj/project.pbxproj
```diff
@@ -445,7 +465,7 @@
...
```
maratal, line 468: And here

Examples/SubscriberExample/Sources/Screens/Map/MapAnnotation/TruckAnnotationView.swift
maratal: @KacperKluka WDYT? Is it always a truck? Or can be a bicycle?

Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/Model/SimulatedLocations.swift
maratal: Oh yeah, I see now, it's simulated!

Examples/SubscriberExample/Sources/Screens/Map/MapAnnotation/TruckAnnotationView.swift
lukasz-szyszkowski: Well, the annotation image view is a truck and this is just an example app. Why do we want to force it to show a bicycle image? Is there any good reason?
So if we want to change the naming, I have to change the image too.

Issue #308, opened 2022-06-06
Publisher.track method doesn't work as described in the code docs


JakubJankowski:
Code documentation for this method says:
```    
If this object was already in this publisher's tracked set then this method only serves to change the actively
tracked object.
```
However, if a trackable object was added earlier by the `Publisher.add()` method, and then I try to `Publisher.track()` that object, then I'm getting an error in the result handler:
```
failure(Trackable with id: *** already exist.)
```
The code docs also say that the `activeTrackable` variable should also be set in this case, but it's still `nil` for me.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-1968) by [Unito](https://www.unito.io)

PR #318, opened 2022-07-26
Feature/308 publisher track method fix

JakubJankowski:
When running the `track` method on a defaultPublisher, if a trackable was added earlier via the `add` method, then the resultHandler would return an error.

This behaviour contradicts the description from the documentation (more info in the [related issue](https://github.com/ably/ably-asset-tracking-swift/issues/308))

Now instead of completing the `track` with an error, we're just setting the active trackable using:

`            enqueue(event: TrackableReadyToTrackEvent(trackable: event.trackable, resultHandler: event.resultHandler))
`


arturjaworski (approved):

KacperKluka (approved):
LGTM :+1:

ikbalkaya (approved):
LGTM 👍

PR #324, opened 2022-08-12
Adjust LocationAnimator Api's methods' and vars' naming so that it's more aligned with the Android sdk

JakubJankowski:


Sources/AblyAssetTrackingUI/LocationAnimator/DefaultLocationAnimator.swift
```diff
@@ -173,11 +178,12 @@ public class DefaultLocationAnimator: NSObject, LocationAnimator {
...
                 $0.endTime >= (link.timestamp - animationStartTime)
             } ?? .zero
-        
             currentAnimationStep = animationSteps[stepIndex]
             
-            // Remove used steps
-            animationSteps.removeSubrange(0..<stepIndex)
+            // Remove used steps and increase the `currentAnimationStepsSinceLastCameraUpdate` count
+            let removeStepsRange = 0..<stepIndex
+            currentAnimationStepsSinceLastCameraUpdate += removeStepsRange.count
```
KacperKluka, line 185: I'm not quite sure how it works on iOS but I wonder if your "animation step" is the same thing that we have :thinking: I'd expect that an animation step is represents a single location from the location update. So if we send camera updates every 5 steps then the camera will move every 5 new locations. Is it how it works? :thinking:

JakubJankowski: Hmm I believe it's the same thing if I understand you correctly ;) 
The animation steps are taken from the `createAnimationStepsFromRequest` method - so those are basically locations from a locationUpdate + skippedLocations, with added info about start and end positions, and later on in the `animationRequestSubject.receive` we add the `endTime` and `duration` to them.

Let me paste in a demo vid so you can easily compare if the result is what you're expecting compared to android:

https://user-images.githubusercontent.com/15122679/184415330-961772a8-a74f-4c53-ac00-d0e2c6701870.mov

KacperKluka (approved):
LGTM :+1:

PR #330, opened 2022-08-17
If we fail to enter presence due to token capabilities, fetch new token and retry

### LINKED ISSUE #311: Request new auth token when channel attach fails due to permission failure
Only for client instances which are using token auth, not basic auth...

When an operation fails against the embedded `AblyRealtime` instance that is indicative of a service-side permissions issue (i.e. capability to perform the operation against the Ably channel is denied) then the SDK should request a new auth token.

We definitely need this for publishers.

We need to investigate whether this needs to also be implemented for subscribers.

See [this Slack comment (internal)](https://ably-real-time.slack.com/archives/C01QU436CSZ/p1655900418554909?thread_ts=1655884714.683019&cid=C01QU436CSZ) for the recorded Zoom discussion between @paddybyers, @KacperKluka, @ikbalkaya, @AndyNicks, @stmoreau and myself. It covers the reasons for taking this approach, plus how it might intersect with a future engineering effort to add an API that would allow a new auth token to be supplied to the publisher instance on addition or removal of a trackable.

This issue has been created for this SDK representing the same functionality outlined in https://github.com/ably/ably-asset-tracking-android/issues/695.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-2083) by [Unito](https://www.unito.io)

lawrence-forooghian:
This closes #311. Please review commit-by-commit; there's lots of information in the commit messages.

For anyone just passing by, it's worth noting that this introduces a new piece of tooling (https://github.com/krzysztofzablocki/Sourcery). Opinions welcome.

### DIFF
--- a/.sourcery-InternalTests.yml
+++ b/.sourcery-InternalTests.yml
@@ -1,0 +1,6 @@
+sources:
+  - Sources/AblyAssetTrackingInternal/AblyWrapper/SDK/AblySDKProtocols.swift
+templates:
+  - Sourcery/templates/AblySDKMocks.stencil
+output:
+  Tests/InternalTests/Mocks/AblySDKMocks.swift
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,3 +86,9 @@ and then manually merge the delta contents in to the main change log (where `v1.
 - SwiftLint is integrated into the project. Make sure that your code does not add any SwiftLint related warning.
 - Please remove default Xcode header comments (with author, license and creation date) as they're not necessary.
 - If you're adding or modifying any part of the public interface of SDK, please also update [QuickHelp](https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_markup_formatting_ref/SymbolDocumentation.html#//apple_ref/doc/uid/TP40016497-CH51-SW1) documentation.
+
+## Generating mocks
+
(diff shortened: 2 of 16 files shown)


lawrence-forooghian:
@KacperKluka - if you're planning on reviewing, probably only the last commit is interesting to you 🙂

lawrence-forooghian:
OK, GitHub doesn't like the fact that I've pushed an 80MB file (the Sourcery binary). Will re-think that approach.

> remote: warning: See http://git.io/iEPt8g for more information.
> remote: warning: File Sourcery/bin/sourcery is 80.84 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
> remote: warning: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.

arturjaworski (approved):

KacperKluka (commented):
I only looked at the `DefaultAbly` changes and they look good :+1: I'm not an iOS dev, therefore I'll leave this as a comment instead of an approve :wink: 

One thing that you might consider adding is an integration test that verifies this behaviour in a real scenario. While developing this feature in the AAT Android, an integration test revealed a problem with `ably-java` that was breaking this feature (local tests were passing but in the CI they were failing). :wink:

lawrence-forooghian:
> One thing that you might consider adding is an integration test that verifies this behaviour in a real scenario.

@KacperKluka Perhaps you didn't see them, but that’s what the tests in `Tests/SystemTests/PublisherAuthenticationSystemTests.swift` and `Tests/SystemTests/SubscriberAuthenticationSystemTests.swift` are doing. They use the real Ably SDK and first supply it with a token with the incorrect capabilities, and then upon retry of token fetch, supply it with a token with the correct capabilities.

PR #320, opened 2022-08-09
Fix docs workflow file structure

KacperKluka:
The docs workflow syntax was broken, now it should be good :wink:

PR #346, opened 2022-08-22
Fix docs workflow YAML

lawrence-forooghian:
This was introduced in b6723ea, which doesn’t seem to have come via a pull request.


lawrence-forooghian:
Well, now it's failing with `Error: Credentials could not be loaded, please check your action inputs: Could not load credentials from any providers`.

lawrence-forooghian:
Just realised this is a duplicate of #320 🤦

Issue #345, opened 2022-08-22
Reuse the "add" logic in the "track" method


KacperKluka:
The `track()` method should perform exactly the same steps as the `add()` one and just add one more step at the end.
Currently, both `add()` and `track()` methods have their own separate implementations, which introduces code duplication. Instead, we should use the `add()` method in the `track()` method implementation. This would eliminate code duplication and make it easier to maintain in the future. So it should look something like this:

```
function track() {
    add()
    finalTrackStep()
}
```

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2600) by [Unito](https://www.unito.io)

KacperKluka:
The whole adding flow can be reused. First, we should fix https://github.com/ably/ably-asset-tracking-swift/issues/340 so the `track()` operation will work correctly when a trackable is already added.

Then the only difference between `add()` and `track()` is that when add completes adding it [calls a callback](https://github.com/ably/ably-asset-tracking-swift/blob/fdc1c4c4f0fd374186e3bc64146b7e0677b3f745/Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift#L344) and the track will [enqueue one more event](https://github.com/ably/ably-asset-tracking-swift/blob/fdc1c4c4f0fd374186e3bc64146b7e0677b3f745/Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift#L240-L245).

For reference check the [Android implementation of both methods](https://github.com/ably/ably-asset-tracking-android/blob/main/publishing-sdk/src/main/java/com/ably/tracking/publisher/CorePublisher.kt#L214-L251).

PR #359, opened 2022-08-26
Reuse the add logic

ikbalkaya:
Code refactor to create a reusable function for add or track
Based on [this comment ](https://github.com/ably/ably-asset-tracking-swift/issues/345#issuecomment-1227070868)
Closes https://github.com/ably/ably-asset-tracking-swift/issues/345

Note to reviewers
I couldn't find a good name and made the function name `performAddOrTrack`. I'd appreciate if you advice a better name.


KacperKluka (approved):
I'm not fluent in Swift but in terms of removing code duplication the PR looks good :+1:

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
```diff
@@ -212,34 +212,8 @@ extension DefaultPublisher {
...
-                        self?.enqueue(event: .trackableReadyToTrack(.init(trackable: event.trackable, resultHandler: event.resultHandler)))
-                    default:
-                        return
-                    }
-                }))
-            case .failure(let error):
-                self?.callback(error: error, handler: event.resultHandler)
-            }
+        performAddOrTrack(event.trackable, resultHandler: event.resultHandler) {
+            self.enqueue(event: .trackableReadyToTrack(.init(trackable: event.trackable, resultHandler: event.resultHandler)))
```
JakubJankowski, line 216: Do we want to enqueue `trackableReadyToTrack` even if `performAddOrTrack` failed?
Earlier we did it only if `presenceJoinedSuccessfully` happened:
```
self?.enqueue(event: .presenceJoinedSuccessfully(.init(trackable: event.trackable) { [weak self] result in
                    switch result {
                    case .success:
                        self?.enqueue(event: .trackableReadyToTrack(.init(trackable: event.trackable, resultHandler: event.resultHandler)))
                    default:
                        return
                    }
                }))
```

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
```diff
@@ -212,34 +212,8 @@ extension DefaultPublisher {
...
-                    case .success:
-                        self?.enqueue(event: .trackableReadyToTrack(.init(trackable: event.trackable, resultHandler: event.resultHandler)))
-                    default:
-                        return
-                    }
-                }))
-            case .failure(let error):
-                self?.callback(error: error, handler: event.resultHandler)
-            }
+        performAddOrTrack(event.trackable, resultHandler: event.resultHandler) {
```
JakubJankowski, line 215: We need to add `[weak self] in` to avoid memory leaks. So this line should be:
```
        performAddOrTrack(event.trackable, resultHandler: event.resultHandler) { [weak self] in
```

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
```diff
@@ -311,36 +285,40 @@ extension DefaultPublisher {
...
                 }))
             case .failure(let error):
-                self?.callback(error: error, handler: event.resultHandler)
+                self?.callback(error: error, handler: resultHandler)
             }
         }
     }
+    // MARK: Add trackable
+    private func performAddTrackableEvent(_ event: Event.AddTrackableEvent) {
+        performAddOrTrack(event.trackable, resultHandler: event.resultHandler) {
```
JakubJankowski, line 318: As above - we need to add `[weak self] in` if we call `self` inside the closure.

JakubJankowski (changes requested):
Generally looks good, just a couple of comments/questions.

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
ikbalkaya: This should have been dealt with inside `performAddOrTrack`

If you check this code block
```swift
 switch result {
            case .success:
                self?.enqueue(event: .presenceJoinedSuccessfully(.init(trackable: trackable) { [weak self] _ in
                    self?.callback(value: Void(), handler: resultHandler)
                }))
            case .failure(let error):
                self?.callback(error: error, handler: resultHandler)
            }
```

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
ikbalkaya: 02def08355771db5938ed30dae6879a75c88b657

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
JakubJankowski: I might be reading this wrong, it's been a long day already, but I think it also won't be invoked when there were no issues and `presenceJoinedSuccessfully`, since you're not calling `completion()` here, only the `callback`:
```swift
                self?.enqueue(event: .presenceJoinedSuccessfully(.init(trackable: trackable) { [weak self] _ in
                    self?.callback(value: Void(), handler: resultHandler)
                }))
```

So currently, if I understand this correctly, the `. trackableReadyToTrack` will be enqueued only if this guard fails (so if a trackable is already added):
```swift
        guard !trackables.contains(trackable) else {
            completion()
            return
        }
```

JakubJankowski: So in essence - I think you should just be calling the completion here instead of a callback:
```swift
self?.enqueue(event: .presenceJoinedSuccessfully(.init(trackable: trackable) { [weak self] _ in
                    completion()
                }))
```

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
JakubJankowski: You'll still need to add a `?` after self, since now you have a weak reference to `self` here:
```
            self?.enqueue(event: .trackableReadyToTrack(.init(trackable: event.trackable, resultHandler: event.resultHandler)))
```
I imagine it's not compiling currently ;)

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
JakubJankowski: As above - a `?` is missing

ikbalkaya: Oops feb724012d3234dceea5ebbb18e3d84fd00ba23b

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
ikbalkaya: feb724012d3234dceea5ebbb18e3d84fd00ba23b

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
ikbalkaya: Of course it's here: 5c4b4986836d7ab92e8755c7f0cc21bed699148a. I'm also brain dead right now but I'd expect there would be. a test coverage for this as this is just refactoring

JakubJankowski (approved):

Issue #336, opened 2022-08-19
Allow to get the list of trackables added to the publisher


KacperKluka:
It should be possible to get a list of the currently added trackables of the publisher. On Android, we've used a `Flow` that emits the trackables set whenever it changes.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2579) by [Unito](https://www.unito.io)

lawrence-forooghian:
On Android, this is the `trackables` flow on `Publisher`; on Swift we'd probably use a new `PublisherDelegate` method.

PR #363, opened 2022-09-06
Allow users to get a list of trackables

JakubJankowski:
Currently it's impossible to know which trackables are tracked by a publisher.
This PR adds a delegate method that returns a list of trackables that are currently added to a publisher whenever this list changes (similarly to how it works on Android).

Resolves: [336](https://github.com/ably/ably-asset-tracking-swift/issues/336)

### DIFF
--- a/Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
+++ b/Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
@@ -42,9 +42,14 @@ class DefaultPublisher: Publisher {
     private var subscribers: [Trackable: Set<Subscriber>]
     private var resolutions: [Trackable: Resolution]
     private var locationEngineResolution: Resolution
-    private var trackables: Set<Trackable>
+    private var trackables: Set<Trackable> {
+        didSet {
+            notifyTrackablesChanged(trackables: trackables)
+        }
+    }
+    
     private var constantLocationEngineResolution: Resolution?
-
+    
     private var lastEnhancedLocations: [Trackable: Location]
     private var lastEnhancedTimestamps: [Trackable: Double]
     private var lastRawLocations: [Trackable: Location]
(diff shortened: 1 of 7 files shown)


KacperKluka (approved):
LGTM in terms of functionality :+1:

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
```diff
@@ -851,6 +856,13 @@ extension DefaultPublisher {
             self.delegate?.publisher(sender: self, didUpdateResolution: event.resolution)
         }
     }
+    
+    private func notifyTrackablesChanged(trackables: Set<Trackable>) {
+        performOnMainThread { [weak self] in
```
ikbalkaya, line 861: Just a question, wouldn't this be a bit more suited to put on code where main thread is performing ? Such as in UI code

ikbalkaya (commented):
Just a question

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
JakubJankowski: Not sure I understand the question, but generally - `performOnMainThread` runs the code on the main, UI thread:
```swift
DispatchQueue.main.async(execute: operation)
```
We're running similar delegate methods on this same thread already as well, like:
notifyDelegateEnhancedLocationChanged,
notifyDelegateConnectionStateChanged,
notifyDelegateResolutionUpdate

etc.

ikbalkaya: I meant that I would expect the delegates  to perform this on main thread but not the `DefaultPublisher` itself. Unless operations are confined on main thread. But I see that this pattern has been used elsewhere in this class so I think it's ok for now.

ikbalkaya (approved):

Issue #347, opened 2022-08-22
Handle the "Connection resume failed" exception from ably-cocoa


KacperKluka:
From time to time we might get the "Connection resume failed" exception from `ably-cocoa` due to the fact that a new connection was established. Because of that, all messages that we were trying to send using the old connection will receive the error. This should not be propagated immediately to the SDK users and it should be handled on the AAT side.

We should try to retry Ably SDK calls if a "connection resume failed" error is received. Here's how it was done on Android https://github.com/ably/ably-asset-tracking-android/pull/665.

Note: There's a known issue in the Android implementation described here https://github.com/ably/ably-asset-tracking-android/issues/744

lawrence-forooghian:
Not yet looked at this one, due to other issues having higher priority.

deanna-lad:
The corresponding Jira issue is https://ably.atlassian.net/browse/SDK-2599

lawrence-forooghian:
This functionality no longer exists in Android as of https://github.com/ably/ably-asset-tracking-android/commit/6542defb68e585338f947cdbbfde64da1f84369e.

PR #364, opened 2022-09-07
Handle connection resume exception

ikbalkaya:
Handles channel connection resume exception by retrying operation one more time

Closes https://github.com/ably/ably-asset-tracking-swift/issues/347

Note to reviewers: I will create a separate task for writing tests for this

### DIFF
--- a/Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
+++ b/Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
@@ -135,8 +135,7 @@ public class DefaultAbly: AblyCommon {
             completion(.success(false))
             return
         }
-        
-        
+
         do {
             try performChannelOperationWithRetry(channel: channelToRemove, operation: { channel in
                 try disconnectChannel(channel: channel, presenceData: presenceData!)
@@ -373,7 +372,6 @@ public class DefaultAbly: AblyCommon {
            try performChannelOperationWithRetry(channel: channel) { channel in
                updatePresenceData(channel: channel, presenceData: presenceData, completion)
             }
-            completion?(.success)
         } catch AblyError.connectionError(let errorInfo){
             completion?(.failure(errorInfo.toErrorInformation()))
         }catch{
@@ -519,7 +517,6 @@ extension DefaultAbly: AblyPublisher {
                     completion?(.success)
                 }
             }
-            completion?(.success)
 
         } catch AblyError.connectionError(let errorInfo){
             completion?(.failure(errorInfo.toErrorInformation()))
--- a/Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
+++ b/Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
@@ -3,7 +3,10 @@ import Ably
 import AblyAssetTrackingCore
 
 public class DefaultAbly: AblyCommon {
-    
+    enum AblyError : Error {
+        case connectionError(errorInfo: ARTErrorInfo)
+    }
+
     public weak var publisherDelegate: AblyPublisherDelegate?
     public weak var subscriberDelegate: AblySubscriberDelegate?
     
@@ -160,6 +163,49 @@ public class DefaultAbly: AblyCommon {
             completion(.failure(error.toErrorInformation()))
         }
     }
+
+    //This function is to be documented and must be invoked from a background thread
+    private func performChannelOperationWithRetry(channel:ARTRealtimeChannel, operation : (ARTRealtimeChannel)->Void) throws {
+      do {
+          logHandler?.w(message: "Trying to perform an operation on a suspended channel \(channel.name), waiting for the channel to be reconnected", error: nil)
+          try waitForChannelReconnection(channel: channel)
+          operation(channel)
+      } catch  AblyError.connectionError(let errorInfo) {
+           if errorInfo.isConnectionResumeException(){
+            
+               logHandler?.w(message: "Connection resume failed for channel \(channel.name), waiting for the channel to be reconnected",
+                       error: errorInfo
+               )
+               //We can try another time
+               do {
+                   try waitForChannelReconnection(channel: channel)
+                   operation(channel)
+               }catch AblyError.connectionError(let secondError){
+                   logHandler?.w(message: "Retrying the operation on channel \(channel.name) has failed for the second time",
+                                          error: secondError
+                                      )
+                   throw AblyError.connectionError(errorInfo: secondError)
+               }
+           }else {
+               throw AblyError.connectionError(errorInfo: errorInfo)
+           }
+      }
+    }
+
+    private func waitForChannelReconnection(channel:ARTRealtimeChannel, timeoutInMs:Int32 = 10_000) throws{
+        guard channel.state.toConnectionState() != .online else{
+            return
+        }
+        let blockingDispatchGroup = DispatchGroup()
+        blockingDispatchGroup.enter()
+        channel.on { stateChange in
+            if (stateChange.current.toConnectionState() == .online){
+                blockingDispatchGroup.leave()
+            }
+        }
+        blockingDispatchGroup.wait(timeout: .now() + Int((timeoutInMs / 1000)))
+    }
+    
     
     public func close(presenceData: PresenceData, completion: @escaping ResultHandler<Void>) {
         let closingDispatchGroup = DispatchGroup()
@@ -465,5 +511,13 @@ extension DefaultAbly: AblyPublisher {
 fileprivate extension ConnectionConfiguration {
     var usesTokenAuth: Bool {
         return authCallback != nil
+    }
+}
+
+extension ARTErrorInfo {
+    func isConnectionResumeException() -> Bool {
+        guard let errorInfo == toErrorInformation() else { return false }
+
+        return  errorInfo.message == "Connection resume failed" && errorInfo.code == 50000 && errorInfo.statusCode == 500
     }
 }
--- a/Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
+++ b/Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
@@ -355,17 +355,30 @@ public class DefaultAbly: AblyCommon {
         }
     }
     
-    public func updatePresenceData(trackableId: String, presenceData: PresenceData, completion: ResultHandler<Void>?) {
-        guard let channel = channels[trackableId] else {
-            return
-        }
-        
+    private func updatePresenceData(channel: AblySDKRealtimeChannel, presenceData: PresenceData, _ completion: ResultHandler<Void>?) {
         channel.presence.update(presenceDataJSON(data: presenceData)) { error in
             if let error = error {
                 completion?(.failure(error.toErrorInformation()))
             } else {
                 completion?(.success)
             }
+        }
+    }
+    
+    public func updatePresenceData(trackableId: String, presenceData: PresenceData, completion: ResultHandler<Void>?) {
+        guard let channel = channels[trackableId] else {
+            return
+        }
+        do{
+           try performChannelOperationWithRetry(channel: channel) { channel in
+               updatePresenceData(channel: channel, presenceData: presenceData, completion)
+            }
+            completion?(.success)
+        } catch AblyError.connectionError(let errorInfo){
+            completion?(.failure(errorInfo.toErrorInformation()))
+        }catch{
+            let artErrorInfo = ARTErrorInfo.create(from: error)
+            completion?(.failure(artErrorInfo.toErrorInformation()))
         }
     }
     
--- a/Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
+++ b/Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
@@ -150,7 +150,7 @@ public class DefaultAbly: AblyCommon {
         }catch AblyError.connectionError(let errorInfo){
             completion(.failure(errorInfo.toErrorInformation()))
         }catch{
-            
+            completion(.failure(ErrorInformation(error: AblyError.connectionError(errorInfo: ARTErrorInfo.create(withCode: 100_000, message: "Unknown error while disconnecting")))))
         }
     }
     
--- a/Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
+++ b/Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
@@ -178,7 +178,7 @@ public class DefaultAbly: AblyCommon {
         }
         
          channel.presence.leave(presenceDataJSON){ errorInfo in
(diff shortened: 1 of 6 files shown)


Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
```diff
@@ -129,38 +132,123 @@ public class DefaultAbly: AblyCommon {
...
+     * Performs the [operation] and if a "connection resume" exception is thrown it waits for the [channel] to
+     * reconnect and retries the [operation], otherwise it rethrows the exception. If the [operation] fails for
+     * the second time the exception is rethrown no matter if it was the "connection resume" exception or not.
+     */
+    private func performChannelOperationWithRetry(channel:AblySDKRealtimeChannel, operation : (AblySDKRealtimeChannel) throws ->Void) throws {
+      do {
+          logHandler?.w(message: "Trying to perform an operation on a suspended channel \(channel.name), waiting for the channel to be reconnected", error: nil)
+          try waitForChannelReconnection(channel: channel)
+          try operation(channel)
+      } catch  AblyError.connectionError(let errorInfo) {
```
KacperKluka, line 204: Why do you call `waitForChannelReconnection()` before doing the operation for the first time? :thinking: Also the log message seems odd here, it looks like some copy-paste error :wink:

KacperKluka (changes requested):

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
```diff
@@ -129,38 +132,123 @@ public class DefaultAbly: AblyCommon {
...
+            channel.unsubscribe()
+            channel.presence.unsubscribe()
+            self.detachFromChannel(channel: channel) { detachError in
+                guard let error = detachError else { return }
+                throw AblyError.connectionError(errorInfo: error)
+            }
+        }
+    }
+    
+    private func leavePresence(channel:AblySDKRealtimeChannel, presenceData: PresenceData?, completion : @escaping (ARTErrorInfo?) throws -> Void){
```
lawrence-forooghian, line 169: I was surprised to see a completion handler with the `throws` modifier. Usually completion handlers are used to inform the caller of a method that the method has completed its work – they're a one-way communication mechanism.

Furthermore, given that the method doesn't ever do anything with any error thrown by `completion`, I don't think that `completion` should have the `throws` modifier.

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
```diff
@@ -129,38 +132,123 @@ public class DefaultAbly: AblyCommon {
...
+        }catch AblyError.connectionError(let errorInfo){
+            completion(.failure(errorInfo.toErrorInformation()))
+        }catch{
+            completion(.failure(ErrorInformation(error: AblyError.connectionError(errorInfo: ARTErrorInfo.create(withCode: 100_000, message: "Unknown error while disconnecting")))))
+        }
+    }
+    
+    private func disconnectChannel(channel:AblySDKRealtimeChannel, presenceData: PresenceData) throws{
+        leavePresence(channel: channel, presenceData: presenceData) { leaveError in
+            if let leaveError = leaveError {
```
lawrence-forooghian, line 156: What's the purpose of throwing this error, given that `leavePresence` ignores it?

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
```diff
@@ -129,38 +132,123 @@ public class DefaultAbly: AblyCommon {
...
+         channel.presence.leave(presenceDataJSON){ errorInfo in
+             //This is to maintain closure signature
+             do{
+                 try completion(errorInfo)
+             }catch{}
+        }
+
+    }
+    
+    private func detachFromChannel(channel:AblySDKRealtimeChannel, completion : @escaping (ARTErrorInfo?) throws -> Void) {
```
lawrence-forooghian, line 186: As in my comment on `leavePresence`, I don't think that `completion` should have the `throws` modifier.

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
```diff
@@ -129,38 +132,123 @@ public class DefaultAbly: AblyCommon {
...
+            if (stateChange.current.toConnectionState() == .online){
+                blockingDispatchGroup.leave()
             }
-            self?.logHandler?.e(message: "\(String(describing: Self.self)): Error while leaving the channel [id: \(trackableId)] presence", error: error)
-            completion(.failure(error.toErrorInformation()))
+            
+        }
+        let timeout: DispatchTime = .now() + .seconds(timeoutInMs / 1000)
+        
+        let result = blockingDispatchGroup.wait(timeout: timeout)
```
lawrence-forooghian, line 244: I haven't looked into it to see which thread this blocks, but I don't think that we'd want to block any thread in order to perform this wait. We could instead use [`DispatchQueue.asyncAfter`](https://developer.apple.com/documentation/dispatch/dispatchqueue/2300020-asyncafter) to run code once the timeout period elapses.

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
```diff
@@ -129,38 +132,123 @@ public class DefaultAbly: AblyCommon {
...
+               throw AblyError.connectionError(errorInfo: errorInfo)
+           }
+      }
+    }
+
+    /**
+     * Waits for the [channel] to change to the [ChannelState.attached] state.
+     * If this doesn't happen during the next [timeoutInMs] milliseconds, then an exception is thrown.
+     */
+    private func waitForChannelReconnection(channel:AblySDKRealtimeChannel, timeoutInMs:Int = 10_000) throws{
```
lawrence-forooghian, line 230: If we're never specifying a value for the `timeoutInMs` parameter, then do we need it?

lawrence-forooghian (commented):
I've added some comments based on a first look at this. I haven't tried to follow the actual flow of what's happening here yet, because I think that it'll be easier for me to do once the things that I mention here have been addressed.

If you'd like to chat about any of my comments, or walk me through what you've done here, drop me a message!

lawrence-forooghian:
@ikbalkaya I was wondering why you created #364? I think that this feature is sufficiently complicated that I would feel much more confident about merging it if we had tests.

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
```diff
@@ -467,3 +572,10 @@ fileprivate extension ConnectionConfiguration {
         return authCallback != nil
     }
 }
+
+extension ARTErrorInfo {
+    func isConnectionResumeException() -> Bool {
+        let errorInfo = toErrorInformation()
+        return  errorInfo.message == "Connection resume failed" && errorInfo.code == 50000 && errorInfo.statusCode == 500
```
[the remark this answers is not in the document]
KacperKluka: Regarding checking the `message` this is, unfortunately, the only way to detect this error 😞 The code and status code are just generic 500s, so we have no better option but to compare the message as well (we do the same [on Android](https://github.com/ably/ably-asset-tracking-android/blob/fd452af5ce76ec1febcd0f6c2885f0e25c0fd795/common/src/main/java/com/ably/tracking/common/Ably.kt#L765-L766))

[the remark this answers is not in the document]
ikbalkaya: I did follow Android's implementation for this 🤷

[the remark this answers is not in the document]
lawrence-forooghian: The `"Connection resume failed"` message seems to be specific to ably-java - [here’s where it’s emitted](https://github.com/ably/ably-java/blob/faa89afa8b6dc4fb535a6f488e3b776c4a0ab8a7/lib/src/main/java/io/ably/lib/transport/ConnectionManager.java#L1121-L1122). That string doesn't appear anywhere in ably-cocoa. Furthermore, in ably-cocoa, it seems that we don't emit any error code of 50000 anywhere.

@KacperKluka do you know if there is there an Ably SDK feature spec point that describes the relevant ably-java behaviour?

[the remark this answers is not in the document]
KacperKluka: Unfortunately, I'm not aware of any specs about this behaviour 😞

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
ikbalkaya: If I'm not mistaken, Swift compiler wouldn't let me pass a throwing closure (function) unless I matched signature of closure using throws. I'll check again and see if I can get rid of this without changing the behaviour

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
ikbalkaya: This is also something I got from Android implementation, but I don't think we need it. I'll remove it.

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
ikbalkaya: > @ikbalkaya I was wondering why you created #366? I think that this feature is sufficiently complicated that I would feel much more confident about merging it if we had tests.

I was about to leave for the pod work so I wanted to leave something reviewable. But now given I am working on this - I'll add tests for this PR

ikbalkaya: 2771dd37dd8400ff10e833e675f3394126ec4c22

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
```diff
@@ -3,7 +3,10 @@ import Ably
 import AblyAssetTrackingCore
 
 public class DefaultAbly: AblyCommon {
-    
+    enum AblyError : Error {
```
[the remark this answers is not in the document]
ikbalkaya: 89d84ae01d1db34106598c9f1a5fa9a62e08e1d7

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
```diff
@@ -129,38 +132,123 @@ public class DefaultAbly: AblyCommon {
...
+            channels.removeValue(forKey: trackableId)
+            completion(.success(true))
+        }catch AblyError.connectionError(let errorInfo){
+            completion(.failure(errorInfo.toErrorInformation()))
+        }catch{
+            completion(.failure(ErrorInformation(error: AblyError.connectionError(errorInfo: ARTErrorInfo.create(withCode: 100_000, message: "Unknown error while disconnecting")))))
+        }
+    }
+    
+    private func disconnectChannel(channel:AblySDKRealtimeChannel, presenceData: PresenceData) throws{
```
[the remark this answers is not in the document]
ikbalkaya: 22e6a062c6a4f1851cbce268470e68715cb379a1

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
ikbalkaya: Wouldn't this error be propagated to the user 🤔

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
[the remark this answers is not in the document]
lawrence-forooghian: I don't think we'll be able to implement this feature on iOS unless we are able to understand what situation triggers this error and how this error is then emitted by the SDK. We'll probably have to start by trying to reproduce this issue on iOS.

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
lawrence-forooghian: As discussed offline let's take a look at this code together

ikbalkaya: I changed the general flow to use completion handler based error propagation rather than using direct throws which was making the whole change messy : 248ff80d6b48c0c988fa637c81c01c9ded685831

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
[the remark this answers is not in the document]
lawrence-forooghian: Cool, I'll take a look into it and let you know.

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
ikbalkaya: This will have been addressed https://github.com/ably/ably-asset-tracking-swift/commit/248ff80d6b48c0c988fa637c81c01c9ded685831

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
ikbalkaya: Also https://github.com/ably/ably-asset-tracking-swift/commit/248ff80d6b48c0c988fa637c81c01c9ded685831

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
ikbalkaya: Also https://github.com/ably/ably-asset-tracking-swift/commit/248ff80d6b48c0c988fa637c81c01c9ded685831

Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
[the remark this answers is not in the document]
ikbalkaya: Spec doesn't really provide any input on re-forming the error message, it just asks for client to emit the error from protocol message. So I think the ably-cocoa implementation is more conforming to the spec.
I think we would probably need input from realtime team about what error code / status code (or maybe message) we should expect on the case. 
I couldn't find a specific one on [this error codes page](https://faqs.ably.com/ably-error-codes)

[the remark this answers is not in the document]
ikbalkaya: According to [this file ](https://github.com/ably/ably-common/blob/b2eeb4e1efa8de83693649314c5d575a096fdb78/protocol/errors.json) the raised error is an internal error 🤔

lawrence-forooghian:
Closing this since #347 is no longer required.

Issue #378, opened 2022-09-27
Revisit the use of milliseconds for update intervals


lawrence-forooghian:
The `LocationAnimator` protocol has the following method:

```swift
func animateLocationUpdate(location: LocationUpdate, expectedIntervalBetweenLocationUpdatesInMilliseconds: TimeInterval)
```

In our SubscriberExample example app, it’s fed from the following method:

```swift
func subscriber(sender: Subscriber, didUpdateDesiredInterval interval: Double) {}
```

where `interval` is in milliseconds.

I think it would make more sense to use the native `TimeInterval` type when referring to durations in our public API, instead of expecting users to remember the units of a particular value.

lawrence-forooghian:
@deanna-lad I noticed that this issue has been closed by Unito sync, do you know why?

deanna-lad:
For some reason there were 2 Jira issues linked to this one github issue so I have closed one of them now as a duplicate. Hopefully the remaining issue will continue to sync.

deanna-lad:
The corresponding Jira issue is https://ably.atlassian.net/browse/SDK-2662

Issue #375, opened 2022-09-22
When using animation, the asset position on map quickly drifts far from received asset positions


lawrence-forooghian:
Whilst investigating #369 I needed to test out the current behaviour of the location animations. I was surprised to find that the position displayed on the subscriber’s map quickly drifts very far from the published positions that it receives.

The full data and methodology is at https://github.com/ably/ably-asset-tracking-swift/compare/main...369-investigate-animation-location-drift, but a graph sums it up well. It shows that, when the publisher is driving down a road, after 140 seconds the map position displayed by the subscriber is around 2km away from the last published location that it received.

![animation-drift](https://user-images.githubusercontent.com/53756884/191769292-d2f50dbe-d53f-4f44-94df-907440d8c078.svg)

This doesn't seem like an intentional behaviour to me. @KacperKluka is this how the Android one behaves?

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2657) by [Unito](https://www.unito.io)

KacperKluka:
On Android we have a constant amount of time for the whole animation that's calculated [here](https://github.com/ably/ably-asset-tracking-android/blob/main/ui-sdk/src/main/java/com/ably/tracking/ui/animation/CoreLocationAnimator.kt#L88-L89). Then we take that time, [divide it](https://github.com/ably/ably-asset-tracking-android/blob/main/ui-sdk/src/main/java/com/ably/tracking/ui/animation/CoreLocationAnimator.kt#L92) by the number of locations we have to animate and [update](https://github.com/ably/ably-asset-tracking-android/blob/main/ui-sdk/src/main/java/com/ably/tracking/ui/animation/CoreLocationAnimator.kt#L93) both the already added and the newly added animation steps to take an equal amount of time. Thanks to this we should not drift away from the current location because the individual animation steps are being sped up and slowed down, so the whole animation fits in the constant animation time.

So the key to not drifting away on Android is to modify the duration of animation steps. Does it also happen on iOS?

Now as I think about it, our approach isn't flawless because we can't modify the step that's currently being animated 🤔 For example, if we have a single animation step that takes 10 seconds, and then while it's animating we receive a new location update the second one will also take 10 seconds as the ongoing step is not on the animation steps list and we can't alter it. But even with that we should not be able to increase this drift endlessly, so it should be acceptable.

lawrence-forooghian:
[...]

PR #379, opened 2022-09-27
Fix some incorrect conversions between milliseconds and seconds in the subscriber example app

lawrence-forooghian:
We had assumed that things that were seconds were milliseconds and vice-versa.

I don’t really understand why we’ve created an API that forces us into this conversion dance, but I’ve created #378 for that.

The updated usage example in the `README` is taken from the SubscriberExample app.

I discovered this whilst investigating #375 (when using animation, the asset position on map quickly drifts far from received asset positions), but this change alone doesn’t fix that issue.

### DIFF
--- a/Examples/SubscriberExample/Sources/Screens/Map/MapViewController.swift
+++ b/Examples/SubscriberExample/Sources/Screens/Map/MapViewController.swift
@@ -30,7 +30,7 @@ class MapViewController: UIViewController {
     private let locationAnimator: LocationAnimator
     private var subscriber: Subscriber?
     private var errors: [ErrorInformation] = []
-    private var locationUpdateInterval: TimeInterval = .zero
+    private var locationUpdateIntervalInMilliseconds: Double = .zero
 
     private var currentResolution: Resolution?
     private var resolutionDebounceTimer: Timer?
@@ -50,7 +50,7 @@ class MapViewController: UIViewController {
         
         self.trackingId = trackingId
         self.locationAnimator = DefaultLocationAnimator()
-        self.locationUpdateInterval = resolution.desiredInterval
+        self.locationUpdateIntervalInMilliseconds = resolution.desiredInterval
 
         let viewControllerType = MapViewController.self
         super.init(nibName: String(describing: viewControllerType), bundle: Bundle(for: viewControllerType))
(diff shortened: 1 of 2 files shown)


Examples/SubscriberExample/Sources/Screens/Map/MapViewController.swift
```diff
@@ -311,6 +311,6 @@ extension MapViewController: SubscriberDelegate {
     }
     
     func subscriber(sender: Subscriber, didUpdateDesiredInterval interval: Double) {
-        locationUpdateInterval = interval
+        locationUpdateInterval = interval / 1000
```
KacperKluka, line 314: Why do we have to do this conversion here? The `interval` is already in milliseconds and we use the `locationUpdateInterval` only in the location animator which expects to get milliseconds. 

What's more, [the initial value](https://github.com/ably/ably-asset-tracking-swift/blob/b3e4817edf63a893ae457757637a14f8560576ed/Examples/SubscriberExample/Sources/Screens/Map/MapViewController.swift#L53) of the `locationUpdateInterval` is set to `resolution.desiredInterval` which is in milliiseconds, but in this line we're saving it in seconds 😅 

I think that we should simply store the `interval` in milliseconds and get rid of the conversion, WDYT? 😉

KacperKluka (changes requested):

Examples/SubscriberExample/Sources/Screens/Map/MapViewController.swift
lawrence-forooghian: I think the conversion was probably done so that we can store the interval as a `TimeInterval`, which is a built-in type for representing durations. But I think you're right, might be neater without the conversion, will change it.

But in #378 I've said that I think we should switch to using `TimeInterval` for durations throughout the SDK.

KacperKluka: If the `TimeInterval` is a standard in iOS I'd say go for it 😉 But on Android, I'm not aware of such a thing, so we would have a slight difference between platforms, but I guess it would be okay, right @QuintinWillison ? 🤔

KacperKluka (approved):
LGTM 👍

Examples/SubscriberExample/Sources/Screens/Map/MapViewController.swift
KacperKluka: Resolving the conversation as it was fixed in https://github.com/ably/ably-asset-tracking-swift/pull/379/commits/5a4db283d68f211dded51bc0e4c531763e02e9ae

PR #380, opened 2022-09-27
Fix a strong reference cycle in `DefaultLocationAnimator`

### LINKED ISSUE #376: Check whether our use of `CADisplayLink` is causing a retain cycle
`CADisplayLink` retains its target, and at a glance it doesn't seem like we've accounted for this fact in `DefaultLocationAnimator`.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2658) by [Unito](https://www.unito.io)

lawrence-forooghian:
Closes #376.

### DIFF
--- a/Sources/AblyAssetTrackingUI/LocationAnimator/DefaultLocationAnimator.swift
+++ b/Sources/AblyAssetTrackingUI/LocationAnimator/DefaultLocationAnimator.swift
@@ -59,6 +59,7 @@ public class DefaultLocationAnimator: NSObject, LocationAnimator {
         }
     }
     private var displayLink: CADisplayLink?
+    private let displayLinkTarget = DisplayLinkTarget()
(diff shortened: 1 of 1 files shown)


maratal (approved):
Seems like `displayLink = nil` in `stopAnimationLoop` would do the job, but I like this independence from calling `stop()`.

lawrence-forooghian:
> Seems like `displayLink = nil` in `stopAnimationLoop` would do the job, but I like this independence from calling `stop()`.

I don't think we need that, since `stopAnimationLoop` already calls [`invalidate`](https://developer.apple.com/documentation/quartzcore/cadisplaylink/1621293-invalidate), which releases the target.

The strong reference cycle we were experiencing was because in our example app we don't call `stop()` (which maybe we should!)

Issue #335, opened 2022-08-19
Allow to enable and disable the cycling profile in the publisher


KacperKluka:
Once the profile is ready we should add an option to enable and disable it in the publisher builder. It should be available as a `VehicleProfile` that can either be `CAR` (the profile disabled) or `BICYCLE` (the profile enabled).

Corresponding Android issue https://github.com/ably/ably-asset-tracking-android/issues/726

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-2580) by [Unito](https://www.unito.io)

lawrence-forooghian:
The default value should be `CAR`.

ikbalkaya:
This is blocked by https://github.com/mapbox-collab/Ably-DH-Collab/issues/8

Issue #256, opened 2022-02-03
Disable predictions in the Mapbox SDK


AndyNicks:
We should allow the users to disable the predictions of Mapbox SDK. Since this can only be set during MapboxNavigation creation we should add a new field in the MapConfiguration object for enabling and disabling predictions. By default, this should be set to true to ensure backwards compatibility.

To disable the predictions we have to set navigatorPredictionMillis to 0.

See https://github.com/ably/ably-asset-tracking-android/pull/547 for additional details

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-1537) by [Unito](https://www.unito.io)

lawrence-forooghian:
The requirements here have changed; now predictions should just be permanently turned off.

We're waiting for a new release of the iOS Mapbox SDK that allows this; we're assuming it will just be a one-line change to our initialization of Mapbox.

ikbalkaya:
Blocked by https://github.com/mapbox-collab/Ably-DH-Collab/issues/1

Issue #377, opened 2022-09-27
The same location is sometimes published twice (once as a `skippedLocation` and once as a `location`)


lawrence-forooghian:
## Summary
Sometimes, the first `skippedLocation` of a published location update is the same as the `location` of the immediately previously-published location update, even when the location of the publisher’s device is continuously moving.

## Where observed
At commit b48e1d1c45e05fd7604c0d76f2bef86e2d4f765c.

## Steps to reproduce
1. Add some logging to `DefaultPublisher.sendEnhancedLocationUpdate` to log the enhanced location update that is published (e.g. see https://github.com/ably/ably-asset-tracking-swift/commit/4544e71ea4cdc4c0e66ed1efd7426a66e1aa3707)
1. Run the `PublisherExampleSwiftUI` app in the iOS Simulator.
1. Set the simulator to use the "Freeway Drive" location profile (Features -> Location).
1. Enter any string as trackable ID, and press "Start publishing".
1. Observe the log output.

## Example log output
```
Sending enhanced location update skippedLocations: [], location: LocationCoordinate(latitude: 37.33521504, longitude: -122.03254905). 
Sending enhanced location update skippedLocations: [AblyAssetTrackingCore.LocationCoordinate(latitude: 37.33521504, longitude: -122.03254905), AblyAssetTrackingCore.LocationCoordinate(latitude: 37.33521504, longitude: -122.03254905), AblyAssetTrackingCore.LocationCoordinate(latitude: 37.33521504, longitude: -122.03254905)], location: LocationCoordinate(latitude: 37.33521504, longitude: -122.03254905). 
Sending enhanced location update skippedLocations: [AblyAssetTrackingCore.LocationCoordinate(latitude: 37.33521504, longitude: -122.03254905), AblyAssetTrackingCore.LocationCoordinate(latitude: 37.33521504, longitude: -122.03254905), AblyAssetTrackingCore.LocationCoordinate(latitude: 37.33521504, longitude: -122.03254905), AblyAssetTrackingCore.LocationCoordinate(latitude: 37.33521504, longitude: -122.03254905)], location: LocationCoordinate(latitude: 37.33521504, longitude: -122.03254905). 
[...]

PR #381, opened 2022-09-28
Cycling profile

ikbalkaya:
This PR enables publishers to set up cycling profile on publisher
Changes
* Add a new VehicleProfile for publisher which has Car and Bicycle in it. Setting this will cause mapbox navigation SDK to set a special profile according to https://github.com/mapbox-collab/Ably-DH-Collab/issues/8#issuecomment-1257837148 . Setting this to `Car` will have no effect - meaning that no profile will be set on Mapbox. I added this to keep things consistent with Android implementation
*  I have made default Vehicle mode [PublisherAuthenticationSystemTests.swift](https://github.com/ably/ably-asset-tracking-swift/pull/381/files#diff-48c0e77b4e4a114ef350cb013cbdc86d471be09b40c50ace845e87ab9407e2a0) to set default mode as `Bicycle`
* Also disabled predictions for Mapbox the instruction for cars too for bicycle mode was to disable it

Closes https://github.com/ably/ably-asset-tracking-swift/issues/335
Also closes https://github.com/ably/ably-asset-tracking-swift/issues/256

I opened https://github.com/ably/ably-asset-tracking-swift/issues/399 as tests were failing here, looks like using class scoped `publisher` caused the tests to fail as some mutation was happening during execution of other tests. I did only move `publisher` and `subscriber` to local test scopes but I think anything that might mutate during a test execution should be moved to local scope later. 
As a result this also fixes https://github.com/ably/ably-asset-tracking-swift/issues/399

### DIFF
--- a/Sources/AblyAssetTrackingPublisher/Publisher.swift
+++ b/Sources/AblyAssetTrackingPublisher/Publisher.swift
@@ -82,5 +82,17 @@ public protocol Publisher {
             - `success` called when the Publisher stopped successfuly.
             - `failure` when an error occurs.
     */
-    func stop(completion: @escaping ResultHandler<Void>) 
+    func stop(completion: @escaping ResultHandler<Void>)
+
+}
+
+public enum VehicleProfile{
+    /**
+        * For cars and motorcycles.
+        */
+    case Car
+    /**
+        * For bicycles and pedestrians.
+        */
+    case Bicycle
 }
--- a/Sources/AblyAssetTrackingPublisher/Publisher.swift
+++ b/Sources/AblyAssetTrackingPublisher/Publisher.swift
@@ -87,12 +87,8 @@ public protocol Publisher {
 }
 
 public enum VehicleProfile{
-    /**
-        * For cars and motorcycles.
-        */
+    /** For cars and motorcycles.*/
     case Car
-    /**
-        * For bicycles and pedestrians.
-        */
+    /** For bicycles and pedestrians.*/
     case Bicycle
 }
--- a/Sources/AblyAssetTrackingPublisher/PublisherBuilder.swift
+++ b/Sources/AblyAssetTrackingPublisher/PublisherBuilder.swift
@@ -67,4 +67,7 @@ public protocol PublisherBuilder {
      Sets an optional `logHandler` that can capture logs from all across the sdk
      */
     func logHandler(handler: AblyLogHandler?) -> PublisherBuilder
+    
+    
+    func vehicleProfile(vehicleProfile: VehicleProfile) -> PublisherBuilder
 }
--- a/Sources/AblyAssetTrackingPublisher/PublisherBuilder.swift
+++ b/Sources/AblyAssetTrackingPublisher/PublisherBuilder.swift
@@ -69,7 +69,7 @@ public protocol PublisherBuilder {
     func logHandler(handler: LogHandler?) -> PublisherBuilder
     
     /**
-     Sets an optional `vehicleProfile`for the publisher, Currently there are two vehicle profiles : Bicycle and Car
+     Sets an optional `vehicleProfile` for the publisher, Currently there are two vehicle profiles : Bicycle and Car
      */
     func vehicleProfile(vehicleProfile: VehicleProfile) -> PublisherBuilder
 }
--- a/Sources/AblyAssetTrackingPublisher/PublisherBuilder.swift
+++ b/Sources/AblyAssetTrackingPublisher/PublisherBuilder.swift
@@ -66,7 +66,7 @@ public protocol PublisherBuilder {
     /**
      Sets an optional `logHandler` that can capture logs from all across the sdk
      */
-    func logHandler(handler: AblyLogHandler?) -> PublisherBuilder
+    func logHandler(handler: LogHandler?) -> PublisherBuilder
     
     
     func vehicleProfile(vehicleProfile: VehicleProfile) -> PublisherBuilder
--- a/Sources/AblyAssetTrackingPublisher/PublisherBuilder.swift
+++ b/Sources/AblyAssetTrackingPublisher/PublisherBuilder.swift
@@ -68,6 +68,8 @@ public protocol PublisherBuilder {
      */
     func logHandler(handler: LogHandler?) -> PublisherBuilder
     
-    
+    /**
+     Sets an optional `vehicleProfile`for the publisher, Currently there are two vehicle profiles : Bicycle and Car
+     */
     func vehicleProfile(vehicleProfile: VehicleProfile) -> PublisherBuilder
 }
--- a/Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
+++ b/Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
@@ -12,19 +12,38 @@ class DefaultLocationService: LocationService {
 
     init(mapboxConfiguration: MapboxConfiguration,
          historyLocation: [CLLocation]?,
-         logHandler: AblyLogHandler?) {
+         logHandler: AblyLogHandler?,
+         vehicleProfile: VehicleProfile) {
         
         let directions = Directions(credentials: mapboxConfiguration.getCredentials())
         NavigationSettings.shared.initialize(directions: directions, tileStoreConfiguration: .default)
+        NavigationSettings.shared.initialize(directions: directions, // or .shared
+                                             tileStoreConfiguration: .default, // or .default
+                                             navigatorPredictionInterval: 0)
         
+        if vehicleProfile == .Bicycle {
+            let cyclingConfig = [
+                        "cache": [
+                            "enableAssetsTrackingMode": true
+                        ],
+                        "navigation": [
+                            "routeLineFallbackPolicy": [
+                                "policy": 1
+                            ]
+                        ]
+                    ]
+            UserDefaults.standard.set(cyclingConfig, forKey: MapboxCoreNavigation.customConfigKey)
+        }
+  
         if let historyLocation = historyLocation {
             replayLocationManager = ReplayLocationManager(locations: historyLocation)
         } else {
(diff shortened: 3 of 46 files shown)


Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
```diff
@@ -12,19 +12,46 @@ class DefaultLocationService: LocationService {
...
+
         if let historyLocation = historyLocation {
             replayLocationManager = ReplayLocationManager(locations: historyLocation)
         } else {
             replayLocationManager = nil
         }
         self.logHandler = logHandler
+        //set location manager with profile identifier only if .Bicycle is provided by clients
+        if vehicleProfile == .Bicycle {
+            self.locationManager = PassiveLocationManager(systemLocationManager: replayLocationManager ,datasetProfileIdentifier: .cycling)
```
KacperKluka, line 50: I don't know Swift so this question might be stupid but I just want to double-check: in the integration guide they are using the below code 
```
let manager = PassiveLocationManager(
            datasetProfileIdentifier: ProfileIdentifier.cycling
        )
```
and there they're using `ProfileIdentifier.cycling`, however, you have just `.cycling` 🤔 Is it the same thing or something different? 🤔

KacperKluka (changes requested):
Looks good but I have a few suggestions 😉

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
```diff
@@ -12,19 +12,46 @@ class DefaultLocationService: LocationService {
...
-        NavigationSettings.shared.initialize(directions: directions, tileStoreConfiguration: .default)
-        
+
+
+        if vehicleProfile == .Bicycle {
+            NavigationSettings.shared.initialize(directions: directions, // or .shared
+                                                 tileStoreConfiguration: .default, // or .default
+                                                 navigatorPredictionInterval: 0,
+                                                 statusPollingConfig: .init(unconditionalPollingPatience: .greatestFiniteMagnitude ,
+                                                                            unconditionalPollingInterval: nil))
```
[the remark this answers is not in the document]
ikbalkaya: There was a different issue about this in this repo :https://github.com/ably/ably-asset-tracking-swift/issues/256
But I think there is no harm doing this in this as it's done for cycling profile 1942d14b37c08367891ba396f76f9ef655ff2a0a

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
ikbalkaya: It's exactly the same

Sources/AblyAssetTrackingPublisher/PublisherBuilder.swift
```diff
@@ -67,4 +67,7 @@ public protocol PublisherBuilder {
      Sets an optional `logHandler` that can capture logs from all across the sdk
      */
     func logHandler(handler: LogHandler?) -> PublisherBuilder
+    
+    
+    func vehicleProfile(vehicleProfile: VehicleProfile) -> PublisherBuilder
```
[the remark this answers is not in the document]
ikbalkaya: Good point 👍 e91a33bbda8f284009d69554b5bd54a718d10815

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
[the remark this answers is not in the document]
KacperKluka: Regarding https://github.com/ably/ably-asset-tracking-swift/issues/256 I think it should be closed when this PR is merged as on Android we removed the option to enable/disable the predictions from the API and we simply always disable it 😉

As I mentioned on Android we are both always disabling predictions and always configuring the status polling on the Mapbox instance. In the https://github.com/ably/ably-asset-tracking-swift/commit/1942d14b37c08367891ba396f76f9ef655ff2a0a you've only disabled predictions, can you also configure the status polling? Overall it should result in moving the whole `NavigationSettings.shared.initialize()` before the `if` 😉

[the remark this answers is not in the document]
ikbalkaya: Yes - I mentioned the closing of the task on PR description
I was on train so I must have missed the comment on status polling configuration. It's done now 39058eb854ef3917979e2c41cf0aa1289b07368a

KacperKluka (approved):
LGTM 👍

Sources/AblyAssetTrackingPublisher/PublisherBuilder.swift
```diff
@@ -67,4 +67,9 @@ public protocol PublisherBuilder {
      Sets an optional `logHandler` that can capture logs from all across the sdk
      */
     func logHandler(handler: LogHandler?) -> PublisherBuilder
+    
+    /**
+     Sets an optional `vehicleProfile`for the publisher, Currently there are two vehicle profiles : Bicycle and Car
```
JakubJankowski, line 72: Super minor comment - between `vehicleProfile` and `for` there's a space missing.

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
```diff
@@ -12,19 +12,43 @@ class DefaultLocationService: LocationService {
...
-        
+         logHandler: LogHandler?,
+         vehicleProfile: VehicleProfile) {
+
         let directions = Directions(credentials: mapboxConfiguration.getCredentials())
-        NavigationSettings.shared.initialize(directions: directions, tileStoreConfiguration: .default)
-        
+
+        NavigationSettings.shared.initialize(directions: directions, // or .shared
+                                             tileStoreConfiguration: .default, // or .default
```
JakubJankowski, line 21: Hmm, I'm a bit confused about what those 2 above comments mean to be honest.

JakubJankowski (approved):
Some very minor comments about... comments, other than that looks good so I'll approve in advance 👍

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
ikbalkaya: Good catch, they were leftover from copy/paste code :) a0f9cb797eef475333d1c1a0955266e622d73eca

Sources/AblyAssetTrackingPublisher/Publisher.swift
```diff
@@ -82,5 +82,17 @@ public protocol Publisher {
...
             - `failure` when an error occurs.
     */
-    func stop(completion: @escaping ResultHandler<Void>) 
+    func stop(completion: @escaping ResultHandler<Void>)
+
+}
+
+public enum VehicleProfile{
+    /**
+        * For cars and motorcycles.
```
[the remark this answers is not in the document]
ikbalkaya: It is a bit off indeed, thanks. 387c37efca13803f3336c12543b267e8fc217151

Sources/AblyAssetTrackingPublisher/PublisherBuilder.swift
ikbalkaya: 387c37efca13803f3336c12543b267e8fc217151

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
ikbalkaya: Resolving on behalf of Jakub

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
```diff
@@ -12,19 +12,43 @@ class DefaultLocationService: LocationService {
...
+         vehicleProfile: VehicleProfile) {
+
         let directions = Directions(credentials: mapboxConfiguration.getCredentials())
-        NavigationSettings.shared.initialize(directions: directions, tileStoreConfiguration: .default)
-        
+
+        NavigationSettings.shared.initialize(directions: directions,
+                                             tileStoreConfiguration: .default,
+                                             navigatorPredictionInterval: 0,
+                                             statusPollingConfig: .init(unconditionalPollingPatience: .greatestFiniteMagnitude ,
```
lawrence-forooghian, line 23: @ikbalkaya Would you please be able to tell me where you got this `statusPollingConfig` line from? It seems to fix #377, which I've written about a bit in 9e935ad1489bba5d5f95291de19889356ef29828. I'd like to perhaps submit it as a separate fix, but would first like to understand this line a bit better.

lawrence-forooghian:
@ikbalkaya as I've [mentioned above](https://github.com/ably/ably-asset-tracking-swift/pull/381/files#r987323322) I might want to steal a bit of the code from this in order to fix #377. I've opened #385 to upgrade MapboxNavigation to the release version of 2.8.0.

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
ikbalkaya: @lawrence-forooghian This was given as an example code usage in this [issue](https://github.com/mapbox-collab/Ably-DH-Collab/issues/8) . If you have access to it, I also asked for explanation for this field. They said

> Predictions will be disabled with navigationPredictionInterval = 0. And setting unconditionalPollingPatience to the maximum value will prevent the routing engine from doing any simulation when there is no signal, such as in a tunnel or urban canyons. This is a parity of InternalUtils.setUnconditionalPollingPatience(Long.MAX_VALUE) in the https://github.com/mapbox-collab/Ably-DH-Collab/issues/3#issuecomment-1203127239

ikbalkaya: Resolving this as this is a separate conversation I had with Mapbox team in https://github.com/mapbox-collab/Ably-DH-Collab/issues/8

PR #382, opened 2022-09-29
Fix `DefaultLocationAnimator`

### LINKED ISSUE #317: Improve animation logic to account delays from CADisplayLInk
Lawrence pointed out that current animation logic depends on CADisplayLink which might fire at different rates and suggested a new approach as below
Start of thread: https://github.com/ably/ably-asset-tracking-swift/pull/289#discussion_r863069102

I think I perhaps didn't explain this very well. What I was saying above is that I think you _can't_ calculate the frames at the start of the animation, because you won't yet know how many frames there are going to be.

As I understand it, you want the animation to last `self.intentionalAnimationDelay + request.interval` seconds (call it `animationDuration`), and to have a constant speed. So I think one way to implement this would be:

- store the time at which the animation begins (call it `startTime`)
- each time the display link fires (call the current time `now`), calculate what proportion of `animationDuration` has passed since the start of the animation: `progress = animationDuration / (now - startTime)`
- figure out which two values you're currently interpolating between: `startIndex = progress * numberOfLocations; endIndex = progress * numberOfLocations + 1` (more or less; check my maths)
- figure out the progress of the animation between those two points: `progressBetweenPoints = progress - startIndex / numberOfLocations`
- and then your next position to animate is `latitude = (locations[endIndex].latitude - locations[startIndex].latitude) * progressBetweenPoints` etc.

_Originally posted by @lawrence-forooghian in https://github.com/ably/ably-asset-tracking-swift/pull/289#discussion_r868112972_

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-2490) by [Unito](https://www.unito.io)

### LINKED ISSUE #369: Check whether CoreLocationAnimator progress may be bigger than 100%
On Android it was possible for the progress to exceed 100% which lead to runtime errors. We should check if that's also the case in this project.

Android issue for reference: https://github.com/ably/ably-asset-tracking-android/issues/751

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2649) by [Unito](https://www.unito.io)

### LINKED ISSUE #375: When using animation, the asset position on map quickly drifts far from received asset positions

lawrence-forooghian:
This fixes various issues related to animation. See commit messages for more details. The bulk of the work lies in the commit "Calculate animation frames on demand".

Closes #317, closes #369, closes #375.

### DIFF
--- a/Sources/AblyAssetTrackingUI/LocationAnimator/DefaultLocationAnimatorCalculator+SupportingTypes.swift
+++ b/Sources/AblyAssetTrackingUI/LocationAnimator/DefaultLocationAnimatorCalculator+SupportingTypes.swift
@@ -1,0 +1,358 @@
+import AblyAssetTrackingCore
+import QuartzCore
+
+extension DefaultLocationAnimatorCalculator {
+    /// Contains the data needed to calculate the events that should be emitted to subscribers of an asset’s map position in the current animation frame. Should be constructed inside a `CADisplayLink` callback invocation.
+    struct Input {
+        /// A moment in time, as described by the return value of `CACurrentMediaTime`.
+        typealias MediaTime = CFTimeInterval
+
+        /// Contains user-specified parameters for configuring the speed of the animation of an asset’s position on a map, and the frequency at which the map is re-centred on the asset’s position.
+        struct Config {
+            /// The number of locations in the input’s `locationsAwaitingAnimation` that the animation must transition between before another camera update will be emitted. Must be non-negative.
+            var locationTransitionsBetweenCameraUpdates: Int {
+                willSet {
+                    precondition(newValue >= 0)
+                }
+            }
+
+            /// An additional duration to be added to the animation duration, on top of the duration given by the input’s `nextLocationUpdatePrediction`. Must be non-negative.
+            var intentionalAnimationDelay: TimeInterval {
+                willSet {
+                    precondition(newValue >= 0)
+                }
+            }
+
+            init(locationTransitionsBetweenCameraUpdates: Int, intentionalAnimationDelay: TimeInterval) {
+                precondition(locationTransitionsBetweenCameraUpdates >= 0)
+                precondition(intentionalAnimationDelay >= 0)
+
+                self.locationTransitionsBetweenCameraUpdates = locationTransitionsBetweenCameraUpdates
+                self.intentionalAnimationDelay = intentionalAnimationDelay
+            }
+        }
+
+        /// Contains information about the state of the world at the start of the current `CADisplayLink` callback invocation.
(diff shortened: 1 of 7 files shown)


ikbalkaya (approved):
I have not checked animation computation in details but code looks good to me 👍

Sources/AblyAssetTrackingUI/LocationAnimator/DefaultLocationAnimatorCalculator+SupportingTypes.swift
```diff
@@ -0,0 +1,358 @@
...
+                        precondition((0 ... 1.0).contains(proportionOfFirstToSecondAlreadyAnimated))
+
+                        self.first = first
+                        self.second = second
+                        self.proportionOfFirstToSecondAlreadyAnimated = proportionOfFirstToSecondAlreadyAnimated
+                        self.remaining = remaining
+                    }
+
+                    /// Like `LocationsAwaitingAnimation`, contains information about the locations of the asset that we still need to animate through. It differs from `LocationsAwaitingAnimation` in that it always contains at least one location; that is, it has no `noLocations` case.
+                    enum OneOrMoreLocationsAwaitingAnimation {
```
maratal, line 134: Shouldn't it be on the same level as `LocationsAwaitingAnimation` (and not embed into it)?

lawrence-forooghian: I thought it made sense to put it into the narrowest scope it needs to be inside. Given that it's only used as a return value of `MultipleLocations.progressing(byNumberOfLocationTransitions:)`, that means inside `MultipleLocations`.

KacperKluka (approved):
I only checked this briefly, it looks more complicated than its Android counterpart, but I assume it's just iOS related stuff 😉 Great that it fixes multiple issues and adds tests for the animator 💪

PR #385, opened 2022-10-04
Upgrade MapboxNavigation to 2.8.0

lawrence-forooghian:
I want to steal a bit of commit b14e3b0 from #381 as it seems to fix issue #377. For that I need the `statusUpdatingSettings` argument of `NavigationSettings.shared.initialize`, which was added in 2.8.0.


KacperKluka (approved):
LGTM 👍

PR #395, opened 2022-10-05
Log locations received from Mapbox and how they change over time

lawrence-forooghian:
I wrote this to help in the investigation of #377, but I think it'll come in useful in the future too so let’s keep it around. See commit messages for more details.

### DIFF
--- a/Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
+++ b/Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
@@ -53,6 +53,7 @@ extension DefaultLocationService: PassiveLocationManagerDelegate {
     }
     
     func passiveLocationManager(_ manager: PassiveLocationManager, didUpdateLocation location: CLLocation, rawLocation: CLLocation) {
+        logHandler?.verbose(message: "\(String(describing: Self.self)), passiveLocationManager(\(manager), didUpdateLocation: \(String(reflecting: location)), rawLocation: \(String(reflecting: rawLocation)))", error: nil)
         delegate?.locationService(sender: self, didUpdateRawLocationUpdate: RawLocationUpdate(location: rawLocation.toLocation()))
         delegate?.locationService(sender: self, didUpdateEnhancedLocationUpdate: EnhancedLocationUpdate(location: location.toLocation()))
     }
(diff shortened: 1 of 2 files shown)


KacperKluka (commented):
Overall it looks good but I'm wondering whether this might impact the battery and introduce some delays in location processing :thinking: Maybe it would be better to add some kind of a debug flag to the SDK and only do those calculations and logging when the flag is enabled? WDYT?

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
```diff
@@ -53,6 +55,17 @@ extension DefaultLocationService: PassiveLocationManagerDelegate {
     }
     
     func passiveLocationManager(_ manager: PassiveLocationManager, didUpdateLocation location: CLLocation, rawLocation: CLLocation) {
+        logHandler?.verbose(message: "\(String(describing: Self.self)), passiveLocationManager(\(manager), didUpdateLocation: \(String(reflecting: location)), rawLocation: \(String(reflecting: rawLocation)))", error: nil)
+
+        if let lastReceivedLocations = lastReceivedLocations {
+            let locationDifferences = LocationDifferenceCalculator.calculateDifferences(first: location, second: lastReceivedLocations.location)
+            logHandler?.verbose(message: "\(String(describing: Self.self)), Differences between location and last received location: \(String(reflecting: locationDifferences))", error: nil)
+            let rawLocationDifferences = LocationDifferenceCalculator.calculateDifferences(first: rawLocation, second: lastReceivedLocations.rawLocation)
+            logHandler?.verbose(message: "\(String(describing: Self.self)), Differences between rawLocation and last received rawLocation: \(String(reflecting: rawLocationDifferences))", error: nil)
```
ikbalkaya, line 64: Shouldn't this and above be `debug` rather than `verbose`?

ikbalkaya (commented):
I think it's a good idea to have the `LocationDifferentCalculator` for debugging, but as far as I can see it's being used regardless of whether the app is in debug mode. 
I also think that we are debugging the behaviour of Mapbox SDK. IMO, it would be cleaner if we get a clean answer from them about the sending of exact locations in different timestamps (I will ask them) and then we can still use this if we really need a utility that computes differences between locations

lawrence-forooghian:
Putting this into draft until we hear back on https://github.com/mapbox-collab/Ably-DH-Collab/issues/16.

lawrence-forooghian:
We don't need this at the moment. Closing.

Issue #389, opened 2022-10-05
Enable to add multiple trackables in the publisher example app


KacperKluka:
The publisher can add multiple trackables which means it can track multiple assets at the same time. However, in the publisher example app, you can only add a single trackable. We should add support for adding multiple trackables, each should have its own ID, resolution settings, and destination. Feel free to check the Android app to see how this is handled in the UI.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2691) by [Unito](https://www.unito.io)

arturjaworski:
Attaching Android example app screenshots

<img src="https://user-images.githubusercontent.com/1076431/194255524-bc9d1363-5a8f-4b3b-bc9a-3186e30d42e1.png" width="50%">
<img src="https://user-images.githubusercontent.com/1076431/194255551-48fef867-5b1c-4fc7-a29b-6e563affb12e.png" width="50%">

lawrence-forooghian:
I think roughly the work will look something like this:

- Have a first page of the app which is the "Create publisher" page. Here you choose all the settings that are used to initialise the publisher (i.e. all of the `PublisherBuilder` methods). We might need to move some of these settings out of the current settings page. Then you press a "Create publisher" button which creates the global publisher.
- Move `Publisher` instance out of `MapViewModel` and into the top level of the app
- Probably want to globally store the last-received information for each trackable (e.g. connection state) so that we have it available when we navigate from the list screen to the map screen (instead of having to wait for another update)
- Create a list of trackables and update it in response to `didChangeTrackables` publisher delegate calls.
- Make this list the first view of the app, and put the existing trackable ID text field and button at the bottom. Change "Start publishing" to "Add", displaying an error if the publisher has not yet been started.
- Create link from trackable list to map screen, which displays (and updates in response to) last-received information for each trackable

I'd imagine the source of truth for all the data being at the list view level, and this data just then being passed in to the map view.

lawrence-forooghian:
I started doing some work on this. Here's the UX that I've got in mind so far. It moves the creation of the publisher to its own screen at the beginning of the app, and puts all of the publisher initialisation-related settings there. Then after creating the publisher, you go to a "publisher details" view, which lists all of its trackables and has a button to add a new trackable.

The "add trackable" screen isn't built yet (it's just the first page of the current version of the app), but it'll let you set things like resolution, destination, etc.

[...]

PR #418, opened 2022-10-13
Choose vehicle profile in the settings screen

JakubJankowski:
Resolves [408](https://github.com/ably/ably-asset-tracking-swift/issues/408)

I made some minor breaking changes to the `VehicleProfile` enum, since (at least in this project) the convention is to use lowercase when naminng enum cases. I also made it a `String` type, and made it conform to `Codable`. Please let me know if you think changes like this are ok.

PR #419, opened 2022-10-13
Allow the user to add multiple trackables in the publisher example app

lawrence-forooghian:
## What does it do?

This is a large change to the publisher example app. It changes the flow so that the user first of all configures and creates a publisher. After creating the publisher, they are able to add as many trackables as they like and view the publisher’s trackables in a list.

This is the majority of the work towards completing #389. I still need to expand the add trackable screen to allow specifying things like resolution.

## Demo

https://user-images.githubusercontent.com/53756884/196203631-c07932ba-39b9-4742-a921-001f3e440e57.mov

(I've fixed the missing Ably logo text in Dark Mode since recording this video.)

Notice that I don't actually try and change the routing profile because it doesn't work. I believe this is an SDK bug, which I'm tracking as #420.

Also notice that some of the trackables stay in the "Connection state unknown" state. I'm tracking this SDK bug as #425.

## Reviewing it

It's a big change but I've tried to structure the commits to make it as reviewable as possible. I'd recommend reviewing commit by commit.


lawrence-forooghian:
I'll fix the conflicts on this (they come from merging #418, so I'll be moving the vehicle profile selection into the create publisher screen).

KacperKluka (commented):
UI-wise it looks and works great 👍
I haven't checked the code, therefore, this is just a "comment" and not an "approve" 😉

arturjaworski (approved):

JakubJankowski (approved):
Generally - looks great! I've not dug super deep into the code, but I didn't see anything obviously wrong.
I've tested it on my device, couldn't find any bugs either :+1:

PR #426, opened 2022-10-17
Add resolution constraints to add trackable screen

### LINKED ISSUE #389: Enable to add multiple trackables in the publisher example app

lawrence-forooghian:
## What it does

Restructures the add resolution screen to use a SwiftUI `Form` and then adds a new set of fields for specifying a resolution constraint, starting with the default values set in the Settings tab.

Closes #389.

(This PR is built on top of #419.)

## Demo

https://user-images.githubusercontent.com/53756884/196253701-763cc86c-71e9-4fc1-a0d2-bccfec066a6c.mp4


KacperKluka (commented):
UI-wise it looks and works great 👍 
I haven't checked the code, therefore, that's a "comment" and not an "approve" 😉

Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/ViewModel/AddTrackableViewModel.swift
```diff
@@ -0,0 +1,50 @@
...
+        guard let accuracy = Accuracy(rawValue: resolutionAccuracy),
+              let minimumDisplacement = Double(resolutionMinimumDisplacement),
+              let desiredInterval = Double(resolutionDesiredInterval) else {
+            return nil
+        }
+        
+        return .init(accuracy: accuracy, desiredInterval: desiredInterval, minimumDisplacement: minimumDisplacement)
+    }
+    
+    func createTrackable() -> Trackable {
```
JakubJankowski, line 35: I don't think this method ever gets called. In `AddTrackableView`, when tapping on the `Add Trackable` button, you're creating a trackable instance with just an id, without using the selected resolution constraints:
```
let trackable = Trackable(id: viewModel.trackableId)
```

lawrence-forooghian: Oops, good catch! Fixed it.

arturjaworski (approved):

PR #435, opened 2022-10-21
Location Source: S3 file

### LINKED ISSUE #290: Replay location data recorded in the common format
We have created a common GeoJSON based format of trip history recordings to be able to replay Android recordings on iOS and vice versa. The JSON schema of locations format can be found [here](https://github.com/ably/ably-asset-tracking-common/blob/main/test-resources/geo/location-schema.json). This is implemented in Android via [LocationHistoryData](https://github.com/ably/ably-asset-tracking-android/blob/main/publishing-sdk/src/main/java/com/ably/tracking/publisher/LocationHistoryModels.kt#L14) and [GeoJsonMessage](https://github.com/ably/ably-asset-tracking-android/blob/main/publishing-sdk/src/main/java/com/ably/tracking/publisher/GeoJsonModels.kt#L8) classes.

The iOS publisher should be able to accept location history data in the common format so they can be replayed (currently it seems that [it only accepts locations in the CLLocation format](https://github.com/ably/ably-asset-tracking-swift/blob/bd26232c269a68ec460f5629d25c1e1783b53e5b/Sources/AblyAssetTrackingPublisher/LocationSource.swift#L12)).

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-1751) by [Unito](https://www.unito.io)

### LINKED ISSUE #410: Allow user to fetch a `LocationHistoryData` file from an S3 bucket and use it as the location source in publisher example app
The user of the publisher example app should be able to fetch a `LocationHistoryData` file from an S3 bucket and use this to simulate the publisher’s current location. See screenshots of the Android implementation: https://github.com/ably/ably-asset-tracking-swift/issues/387#issuecomment-1269520506

## Depends on

We first of all need to:

- add to the AAT SDK the ability to use a `LocationHistoryData` object as a location source (#290)
- add to the publisher app a class for fetching file lists and `LocationHistoryData` files from an S3 bucket (#411)

## UI

The user of the publisher example app should be able to choose which location source the publisher should use - that is, what value should be passed to `PublisherBuilder.locationSource(_:)`. They should have the option of choosing the default - i.e. to not call `PublisherBuilder.locationSource(_:)` (which will cause the SDK to use the device’s location).

If the user chooses the `LocationHistoryData` location source, then they will need to choose the data that should be used. The app will load a list of files from the S3 bucket and the user must choose one. The app will then download and parse the file and use it when building the publisher.

The chosen value should be persisted, like the other settings in the app.

Ideally, this would be added to the "Create publisher" screen [that I'm adding for #389](https://github.com/ably/ably-asset-tracking-swift/issues/389#issuecomment-1274577829). If that's not yet done, then add it to the app's existing Settings screen and I'll move it afterwards.

## Fallback

~The ability to use S3 is an optional feature and requires an S3 configuration to be provided. If it's not provided, then the app should not allow the user to select the `LocationHistoryData` location source and display some sort of useful message explaining why it's not available.~ Moved to #436 

## Documentation

~We should document in the example apps section of the `README` how S3 can be used and how to configure it.~ Moved to #437

## Original issue

Split from #387.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2711) by [Unito](https://www.unito.io)

### LINKED ISSUE #411: Add class to publisher example app for fetching file list and `LocationHistoryData` objects from an S3 bucket
This would be an equivalent of the [`S3Helper`](https://github.com/ably/ably-asset-tracking-android/blob/main/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/S3Helper.kt) class in the Android publisher example app. We don't need the ability to upload data at the moment, though; we'll do that when we need to in #414.

The `LocationHistoryData` type will be added to our SDK in #290.

That class wraps the [Amplify](https://docs.amplify.aws) framework – an Amazon framework for interacting with AWS. Amplify [is also available on iOS](https://docs.amplify.aws/lib/q/platform/ios/), so let's use that unless there's a good reason not to. There are two versions of Amplify available for iOS, one of which is in developer preview and "exclusively using Swift". Use whichever seems easiest to work with.

The input for this class is some sort of configuration for Amplify - on Android they use an `amplifyconfiguration.json` file bundled with the app. (Ably already has an S3 bucket set up with some example files in it – if you'd like an `amplifyconfiguration.json` example for accessing this bucket, contact @lawrence-forooghian or @KacperKluka.) Bear in mind that this file is optional - some users might not have it, in which case they should be able to use the publisher app without using the S3 functionality.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2712) by [Unito](https://www.unito.io)

### LINKED ISSUE #436: S3 configuration should be optional
> The ability to use S3 is an optional feature and requires an S3 configuration to be provided. If it's not provided, then the app should not allow the user to select the LocationHistoryData location source and display some sort of useful message explaining why it's not available.

Moved from #410 

Also amplifyconfiguration.json (now it's empty) should be removed from repository to avoid commit actual configuration file.

arturjaworski:
Closes #410, closes #290, closes #411, closes #436

Created #437 for documentation (@lawrence-forooghian will address)


KacperKluka (commented):
I've not looked into the code as I'm not an iOS developer, but regarding the functionality it looks good 👍

Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/Helper/S3Helper.swift
```diff
@@ -0,0 +1,43 @@
...
+class S3Helper {
+        
+    init() {
+        do {
+            try Amplify.add(plugin: AWSCognitoAuthPlugin())
+            try Amplify.add(plugin: AWSS3StoragePlugin())
+            try Amplify.configure()
+            print("Amplify configured with Auth and Storage plugins")
+        } catch {
+            print("Failed to initialize Amplify with \(error)")
```
JakubJankowski, line 16: A note to all the prints in this PR - we have a `PublisherLogger` that we use in the publisher factory method in `CreatePublisherViewModel`. We should probably use it instead of those prints (for example by using it as a singleton).

arturjaworski: done

JakubJankowski:
One thing that's missing is some info in the `Readme.md` `## Example Apps` section on how to use this feature. What's the `amplifyconfiguration` file? Where can users get it from? etc.

It doesn't have to be a part of this PR, but in that case I'd create a separate issue for this and handle it rather soon-ish.

arturjaworski:
> One thing that's missing is some info in the `Readme.md` `## Example Apps` section on how to use this feature. What's the `amplifyconfiguration` file? Where can users get it from? etc.
> 
> It doesn't have to be a part of this PR, but in that case I'd create a separate issue for this and handle it rather soon-ish.

Its already created #437

JakubJankowski (approved):
LGTM 👍

arturjaworski:
One CI action is failing, but it's looks like Xcode bug. New Xcode should be released soon with iOS 16.1, so I'm not does it make sense to fix it.

lawrence-forooghian:
I'll take a look at this today, thanks!

lawrence-forooghian:
I'll review this PR and take responsibility for any changes needed, now that @arturjaworski is no longer on the project.

I'm assuming the issue mentioned in the PR description is meant to be #290, not 230 (which is very unrelated). Have updated PR description.

lawrence-forooghian:
There's nothing in the commits or PR description to explain why we've changed Xcode version / simulator. I'll switch that back to see what I can find out.

lawrence-forooghian:
OK, I'm assuming that the update was due to this error:

```
xcodebuild: error: Could not resolve package dependencies:
[6931](https://github.com/ably/ably-asset-tracking-swift/actions/runs/3320947815/jobs/5487945236#step:6:6932)
  Dependencies could not be resolved because 'amplify-swift' >= 2.0.0 contains incompatible tools version (5.6.0) and root depends on 'amplify-swift' 2.0.0..<3.0.0.
```

lawrence-forooghian (approved):

PR #439, opened 2022-10-25
Update to using Xcode 13.4.1 in CI

lawrence-forooghian:
We want to be able to use the amplify-swift package, which requires Swift >= 5.6.

I haven’t done a major version update (i.e. to Xcode 14) since when Artur tried that in PR #435 he seemed to be getting some errors that I don’t want to investigate right now.


lawrence-forooghian:
I'm going to go ahead and merge this one since I think it's quite uncontroversial, nobody is around on AAT at the moment, and it will unblock #435. Please add comments retroactively if you have any and I'll address them. Thanks!

Issue #342, opened 2022-08-22
Handle a scenario where a user tries to remove a trackable that's still being added


KacperKluka:
There can be a situation in which a trackable is being added and a moment later it is being removed. That's because both `add()` and `remove()` operations are async ones. What happens now is not predictable as the result depends on whether the trackable adding operation was completed before the removal began. What should happen is that the trackable should always be removed if someone called the `remove()` method. So if the trackable that's requested to be removed is currently in the add process we should stop that process, revert all operations that were performed until now and then notify that the removal was successful.

On Android, we have a [`TrackableRemovalGuard`](https://github.com/ably/ably-asset-tracking-android/blob/main/publishing-sdk/src/main/java/com/ably/tracking/publisher/guards/TrackableRemovalGuard.kt) that's used to mark trackables for removal if they're still during the add process. Then each step of the add process checks if the trackable was marked for removal and if it was they revert the add process.

lawrence-forooghian:
This should be done alongside #341.

lawrence-forooghian:
@KacperKluka What should be the result of the in-progress add operation? Should it fail?

KacperKluka:
Yes, the in-progress add operations should fail with a special exception ([here's](https://github.com/ably/ably-asset-tracking-android/blob/main/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/TrackableRemovalRequestedWorker.kt#L23-L26) how it works on Android).

deanna-lad:
The corresponding Jira issue is https://ably.atlassian.net/browse/SDK-2594

Issue #341, opened 2022-08-22
Handle a scenario where the same trackable is being added multiple times


KacperKluka:
Since we work in a multi-threaded environment we should properly handle a scenario where the same trackable is being added at the same time from multiple places. Currently in that case each add operation would call all methods and try to connect to Ably channels etc. What should happen is the trackable adding steps should be called only once and all the other add operations should simply wait for the result of the first one.

On Android, we handle the situation with the help of the [`DuplicateTrackableGuard`](https://github.com/ably/ably-asset-tracking-android/blob/main/publishing-sdk/src/main/java/com/ably/tracking/publisher/guards/DublicateTrackableGuard.kt) that makes sure only one add operation is being executed. All other add calls will be added to a list and notified when the first operation finishes and what was the result of that operation.

lawrence-forooghian:
Just to explain how this is different to https://github.com/ably/ably-asset-tracking-swift/issues/340. Adding a trackable is an asynchronous operation that can last several seconds. This issue relates to the behaviour if you try and add a trackable _whilst an add operation for a trackable with the same ID is still in progress_, whereas #340 is about if you try and add a trackable when one with the same ID has already completed being added.

lawrence-forooghian:
This should be done alongside #341 as they'll probably require the creation of similar mechanisms.

lawrence-forooghian:
I'm unassigning this from myself since I'm not actively working on it. My work-in-progress can be found at https://github.com/ably/ably-asset-tracking-swift/compare/main...341-same-trackable-being-added-multiple-times (it's based on top of #440).

deanna-lad:
The corresponding Jira issue is https://ably.atlassian.net/browse/SDK-2595

PR #440, opened 2022-10-27
Make a few readability-related refactors to `DefaultPublisher`

lawrence-forooghian:
This is some groundwork that I did a while ago for #341 and #342. Thought I might as well get it in now. Nothing very controversial here, I think. See commit messages for more details.


maratal (approved):
LGTM

PR #444, opened 2022-11-04
Deprecate ability to specify client ID when using token-based auth

### LINKED ISSUE #384: Make the client ID an optional parameter if users use a token-based auth
When users use a token-based auth we should make the client ID an optional configuration parameter.

For more details look here https://github.com/ably/ably-asset-tracking-common/issues/59

lawrence-forooghian:
Closes #384.


KacperKluka (approved):
LGTM 👍

Issue #445, opened 2022-11-07
Clarify API rules re which from which threads the SDK can be interacted with


lawrence-forooghian:
WIP, will fill in details

PR #446, opened 2022-11-07
Publisher example: Make sure we start publisher from main thread

lawrence-forooghian:
I noticed that in the publisher example app, the trackables are no longer transitioning to the online state. This appears to have been broken by a change in ac20806 where we started creating the publisher inside an `async` function and hence potentially no longer on the main thread.

This SDK does not explain the rules re from which threads it can be interacted with. I have created #445 to decide and document these rules.


JakubJankowski (approved):
LGTM

PR #457, opened 2022-11-10
Publisher example app: Upload location history data to S3 after emitted by publisher SDK

### LINKED ISSUE #111: Add support for recording location history data in the Publisher
On iOS, there's no support for recording the trip data. You can only replay the data that was recorded on Android. We should be able to collect the location history data on iOS too.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-737) by [Unito](https://www.unito.io)

### LINKED ISSUE #413: Upload location history data to S3 after it's emitted by the SDK in the publisher example app
h2. Description
In the Android publisher example app, if the app has been given an S3 configuration, then whenever the asset tracking SDK emits a {{locationHistory}} callback (i.e. when it says that some trip location recorded data is available) [then the app will upload this data to S3|https://github.com/ably/ably-asset-tracking-android/blob/8a9d6b99b45d3a41d333ff0d60337ee79f40c4a8/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/PublisherService.kt#L86-L90].

We want the iOS app to do the same thing.

h2. Depends on
* #410 - we need the ability to provide the publisher app with an S3 config
* #414 - we need the ability to use our S3 client to upload {{LocationHistoryData}} objects to S3
* #111 - we need the {{locationHistory}} callback to be implemented in the publisher

### LINKED ISSUE #414: Add ability to upload data to S3 in publisher app S3 client class
This adds a new method to the S3 client class introduced in #411, to allow the user to upload a `LocationHistoryData` object to S3. See [the Android `uploadHistoryData`](https://github.com/ably/ably-asset-tracking-android/blob/main/publishing-example-app/src/main/java/com/ably/tracking/example/publisher/S3Helper.kt#L97-L117) method.

lawrence-forooghian:
**Note this is based on #471, so please review that one first.**

Closes #413. See commit messages for more details.

It adds a UI for viewing the status of the uploads and the names of the uploaded files. Here's a video of the UI:

https://user-images.githubusercontent.com/53756884/203398937-1a34e031-b2ce-4777-afc8-b5d78f54d105.mov


JakubJankowski (approved):

KacperKluka (approved):
LGTM 👍

Issue #420, opened 2022-10-13
When we try to change the routing profile, the operation never completes


lawrence-forooghian:
The SDK appears to never be calling the callback of `Publisher.changeRoutingProfile(profile:, completion:)`. For example, when using the ["Change routing profile" button in the publisher example app](https://github.com/ably/ably-asset-tracking-swift/blob/d597c0252a69d8e5b3b81dc0bc1cbecb2c261372/Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/View/PublisherDetailsView.swift#L57-L96).

https://user-images.githubusercontent.com/53756884/196193637-0b71dea9-5230-45f0-924b-258c5c2ecac8.mov

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2721) by [Unito](https://www.unito.io)

ikbalkaya:
This bug seems to be caused partly by us partly by Mapbox that enforces a destination when providing route options. See
https://github.com/ably/ably-asset-tracking-swift/blob/2a41b54585d5eb84dcf53d2326d82fbe9f4d6fc2/Sources/AblyAssetTrackingPublisher/Services/RouteProvider.swift#L59

Our code just ignores result handler when destination is set to null - so I provided a temporary fix that will at least invoke result handler. But in Android destination [seems to be optional ](https://github.com/ably/ably-asset-tracking-android/blob/c82d8d3cdaf681396940387d6a19928a96eb68a5/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ChangeRoutingProfileWorker.kt) (@KacperKluka to confirm)

So we will either need to provide some arbitrary destination value or have to ask Mapbox for this

KacperKluka:
The routing profile is tightly coupled with the destination. The only place where we specify a routing profile in Mapbox is while setting routes. In order to set a route we need a destination. So we need to have both destination and routing profile to properly set a route. 

However, it's not required to provide a destination for a trackable as you've noticed [here](https://github.com/ably/ably-asset-tracking-android/blob/c82d8d3cdaf681396940387d6a19928a96eb68a5/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ChangeRoutingProfileWorker.kt#L14). So when there is no destination we simply don't set a route. I'm not exactly sure what's the issue here but it's probably an AAT problem 😉

ikbalkaya:
[...]

PR #458, opened 2022-11-10
Partial fix for routing profile change issue

ikbalkaya:
This is a partial fix for  https://github.com/ably/ably-asset-tracking-swift/issues/420

The problem was that, the code paths wouldn't invoke result handler in all situations. For example when destination was nil - as in this case, the function would just return without bothering with result handler.
Here I made a change that would make sure that result handler is invoked when destination is nil with a failing response when change profile action is performed.

But In Android [destination is optional.](https://github.com/ably/ably-asset-tracking-android/blob/c82d8d3cdaf681396940387d6a19928a96eb68a5/publishing-sdk/src/main/java/com/ably/tracking/publisher/workerqueue/workers/ChangeRoutingProfileWorker.kt)
This solves only the issue of us not providing a result handler when destination is nil. But the important bug is that Mapbox does not provide a way to pass an optional destination 

See https://github.com/ably/ably-asset-tracking-swift/issues/420#issuecomment-1311567241

### DIFF
--- a/Sources/AblyAssetTrackingPublisher/Services/RouteProvider.swift
+++ b/Sources/AblyAssetTrackingPublisher/Services/RouteProvider.swift
@@ -3,7 +3,7 @@ import MapboxDirections
 import AblyAssetTrackingCore
 
 protocol RouteProvider {
-    func getRoute(to destination: CLLocationCoordinate2D, withRoutingProfile routingProfile: RoutingProfile, completion: @escaping ResultHandler<Route>)
+    func getRoute(to destination: CLLocationCoordinate2D?, withRoutingProfile routingProfile: RoutingProfile, completion: @escaping ResultHandler<Route>)
     func changeRoutingProfile(to routingProfile: RoutingProfile, completion: @escaping ResultHandler<Route>)
 }
 
@@ -23,12 +23,7 @@ class DefaultRouteProvider: NSObject, RouteProvider {
 
     func changeRoutingProfile(to routingProfile: RoutingProfile, completion: @escaping ResultHandler<Route>) {
         self.routingProfile = routingProfile
-        guard let destination = destination else {
-            let errorInformation = ErrorInformation(type: .publisherError(errorMessage: "Cannot change routing profile when destination is not available"))
-            completion(.failure(errorInformation))
-            return
-        }
-        
+    
         if isCalculating(resultHandler: completion){
             return
         }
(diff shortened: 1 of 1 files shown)


Sources/AblyAssetTrackingPublisher/Services/RouteProvider.swift
```diff
@@ -23,8 +23,13 @@ class DefaultRouteProvider: NSObject, RouteProvider {
 
     func changeRoutingProfile(to routingProfile: RoutingProfile, completion: @escaping ResultHandler<Route>) {
         self.routingProfile = routingProfile
-        guard let destination = self.destination,
-              !isCalculating(resultHandler: completion) else {
+        guard let destination = destination else {
+            let errorInformation = ErrorInformation(type: .publisherError(errorMessage: "Cannot change routing profile when destination is not available"))
+            completion(.failure(errorInformation))
```
KacperKluka, line 28: I'm not sure if we should say that the `changeRoutingProfile` operation fails because there is no trackable with a destination active in the publisher at the moment 🤔 You should be able to change the routing profile even if you do not have a single trackable added. I'd say that if there is no destination then you simply change the routing profile and treat it as a success 🤔

KacperKluka (changes requested):

Sources/AblyAssetTrackingPublisher/Services/RouteProvider.swift
ikbalkaya: @KacperKluka In Android example app, this seems to be set only when a publisher is created which is not the case here. I wonder why we do have this option of changing this after creating the publisher on iOS app. In the first page there is already an option of Routing profile which only changes a setting. I'm also not sure about changing the result interface for this - if it's a success it will expect a route which is really not the case here. Is there a spec on what this changing profile button should do @lawrence-forooghian ?

lawrence-forooghian: > Is there a spec on what this changing profile button should do

As far as I'm aware, there aren't any formal requirements for the example apps, but perhaps @KacperKluka knows otherwise.

> I wonder why we do have this option of changing this after creating the publisher on iOS app.

In my mind, the idea of the publisher example app is to expose the full API of the publisher SDK. The `Publisher.changeRoutingProfile(profile:,completion:)` method doesn’t document any restrictions on when it’s okay to call it, so in the example app I allow the user to call it whenever they want.

Either we need to document restrictions on when this method can be called, or we need to allow it to always be called and make sure the callback is always called.

lawrence-forooghian: > I'd say that if there is no destination then you simply change the routing profile and treat it as a success 🤔

Is that how it behaves on Android, @KacperKluka? Our aim here should be to be consistent.

lawrence-forooghian:
Hey @ikbalkaya, I'll address Kacper's comments here since you're working on other stuff!

lawrence-forooghian:
Closing this in favour of https://github.com/ably/ably-asset-tracking-swift/pull/474.

PR #459, opened 2022-11-15
Fix bug where, after removing a trackable, can’t add another with same ID

### LINKED ISSUE #431: Can’t re-add a trackable with same ID after removing it from publisher
## Steps to reproduce

This can easily be reproduced in the publisher example app.

1. Add a trackable with ID 123.
2. Remove the trackable.
3. Try to add a new trackable with ID 123.

## Expected outcome

The second trackable with ID 123 is added successfully.

## Actual outcome

The second trackable with ID 123 cannot be added, failing with error `unable to enter presence channel (incompatible channel state: Detached)`.

## Demo

https://user-images.githubusercontent.com/53756884/196274402-aa0b0587-7118-4962-85fd-5f30283c98ae.mov

lawrence-forooghian:
This occurs because the second time around, the channel will be in a `DETACHED` state, and trying to enter presence on that channel will give the error `unable to enter presence channel (incompatible channel state: Detached)`.

So, we follow [Android’s approach](https://github.com/ably/ably-asset-tracking-android/blob/d61ca7f3a1dc2f8c3ed5eb5def21ee9a3e0229b8/common/src/main/java/com/ably/tracking/common/Ably.kt#L363-L365), and check the channel’s state and attach if necessary before we try to enter presence.

Closes #431.

### DIFF
--- a/Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
+++ b/Sources/AblyAssetTrackingInternal/AblyWrapper/DefaultAbly.swift
@@ -68,6 +68,28 @@ public class DefaultAbly: AblyCommon {
         
         let channel = client.channels.getChannelFor(trackingId: trackableId, options: options)
         
+        if [.detached, .failed].contains(channel.state) {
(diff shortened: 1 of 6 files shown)


JakubJankowski (approved):
LGTM 👍

Issue #447, opened 2022-11-09
Add example apps to App Store Connect


JakubJankowski:
There's a need to test the iOS example apps internally asap without every tester having to compile the apps themselves. 
We'll be using TestFlight as a distribution platform, so we'll need to:

1. Create App Store Connect apps/projects for both the `PublisherExampleSwiftUI` and `SubscriberExample` apps. Please consult with @QuintinWillison before selecting the app names, SKUs, bundle ids etc.
2. "Bake in" the Ably API key and s3 credentials, since those builds won't be available publicly 
3. Upload the first builds to TestFlight manually. 

[Here](https://ably-real-time.slack.com/archives/C01EPJENRD0/p1667915221054679) is a related slack discussion.

PR #460, opened 2022-11-16
Prepare the project's example apps for TestFlight distribution

JakubJankowski:
Resolves [447](https://github.com/ably/ably-asset-tracking-swift/issues/447)

I've added a new set of app icons (can't upload to TestFlight without assigning the required icons, but might as well just add all of them).

Changed the example apps' bundle ids and Display Names, so that they are now consistent with the ones on Android.

I also added a new entry to the Info.plist files, so that we don't have to manually confirm the export compliance in TestFlight every time a new build lands:
```
	<key>ITSAppUsesNonExemptEncryption</key>
	<false/>
```

The first builds are already on TestFlight.


Examples/SubscriberExample/Resources/Info.plist
```diff
@@ -2,8 +2,12 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>ABLY_API_KEY</key>
```
lawrence-forooghian, line 5: What is this `ABLY_API_KEY`? It doesn't seem related to anything else in this PR.

lawrence-forooghian: Ah, I see it's just moved.

lawrence-forooghian (approved):

Issue #455, opened 2022-11-10
Fix intermittent connection issue on publisher


ikbalkaya:
To fix intermittent connection issue on publisher,

As a solution, @paddybyers wrote [internal slack message](https://ably-real-time.slack.com/archives/C03F61P99S6/p1668093705229429?thread_ts=1668093554.949789&cid=C03F61P99S6)

> AAT needs to set a remainPresentFor transportParam in the ClientOptions when it instances the Ably library.
> Optionally it should expose the ability to configure that to the AAT consumer

PR #462, opened 2022-11-17
Allow setting `remainPresentFor` in Publisher's ConnectionConfiguration

JakubJankowski:
Resolves [455](https://github.com/ably/ably-asset-tracking-swift/issues/455) by allowing users to set the `remainPresentFor` property on a ConnectionConfiguration object.

Should allow users to fix an issue when a rider enters to or exits from a channel intermittently and is still aware of rider presence continually.


Sources/AblyAssetTrackingInternal/Configuration/ConnectionConfiguration+Extensions.swift
```diff
@@ -27,6 +27,11 @@ extension ConnectionConfiguration {
             clientOptions.environment = environment
         }
         
+        if let remainPresentForMilliseconds = remainPresentForMilliseconds {
```
lawrence-forooghian, line 30: Could we add a test to `Tests/InternalTests/Configuration/ConnectionConfiguration+ExtensionsTests.swift` which tests that it correctly sets the `transportParams` when `remainPresentInMilliseconds` is specified?

lawrence-forooghian (commented):
Looks good, just one comment.

Sources/AblyAssetTrackingInternal/Configuration/ConnectionConfiguration+Extensions.swift
JakubJankowski: Good point, added in 06526ce6fd20524221d48851b6b26da65f80026c

JakubJankowski: Good point, added in 06526ce6fd20524221d48851b6b26da65f80026c

lawrence-forooghian (approved):

PR #467, opened 2022-11-22
Handle versioning of archived `LocationHistoryData` data

### LINKED ISSUE #466: Handle versioning of archived `LocationHistoryData` data
Handle the versioning rules introduced in https://github.com/ably/ably-asset-tracking-common/pull/67, when unarchiving `LocationHistoryData` objects.

- Choose which parser to use based on the `version` field (i.e. use version 1 if the version is number `1`, and if version is a semantic version string then use its major version number)
- Raise an error if the `version` field has an unsupported major version number.
- I also think we should remove (well, deprecate) the `version` field from the public interface of the `LocationHistoryData` class. It's an implementation detail of how the data is stored. Let's deprecate it and make it just always return `1`.
- Set the `version` field correctly when archiving.

lawrence-forooghian:
Closes #466. See commit messages for more details.

### DIFF
--- a/Examples/AblyAssetTracking.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Examples/AblyAssetTracking.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -326,6 +326,15 @@
         }
       },
       {
+        "package": "Version",
+        "repositoryURL": "https://github.com/mxcl/Version",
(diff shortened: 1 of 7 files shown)


KacperKluka (approved):
I don't have much Swift knowledge so I have not assessed the code. But the overall solution looks well-thought and working as intended 👍

Issue #456, opened 2022-11-10
Write tests for `DefaultLocationService`


lawrence-forooghian:
This class is currently untested. We should test it. To do so will require us to mock out the Mapbox SDK.

PR #471, opened 2022-11-22
Record location history data and expose it in publisher API

### LINKED ISSUE #111: Add support for recording location history data in the Publisher

lawrence-forooghian:
The implementation is roughly taken from Mapbox.kt of the Android asset tracking codebase at d61ca7f.

This increases the complexity of DefaultLocationService, and I would have liked to have written tests for this new functionality, but the class currently has no tests and it would require a chunk of work to make it testable (we need to mock out Mapbox). So I’ve created a separate issue #456 for writing tests for this class.

Closes #111.


JakubJankowski (approved):
Looks good and makes perfect sense 👍

KacperKluka (approved):
In terms of functionality LGTM 👍

PR #474, opened 2022-11-22
Fix situations where `changeRoutingProfile` doesn’t call its callback

### LINKED ISSUE #420: When we try to change the routing profile, the operation never completes

lawrence-forooghian:
In situations where there either was no active trackable, or the active trackable had no destination, the completion handler of `changeRoutingProfile(profile:completion:)` would not be called.

This fixes that by making sure that we always call the callback. In the case where there is no active trackable with a destination, we simply update the publisher’s routing profile and call the callback immediately. This is consistent with the behaviour on Android, [as described by Kacper](https://github.com/ably/ably-asset-tracking-swift/pull/458#discussion_r1029048629).

Our existing tests for this method didn’t catch this because they weren’t actually waiting for the callback, and so none of their assertions were executed. So I’ve fixed those tests and added some more.

The fact that the tests’ assertions weren’t being called also hid another issue with `DefaultPublisher` – namely, that it was ignoring the routing profile that was passed to `changeRoutingProfile`. I’ve fixed this, too.

Closes #420.

### DIFF
--- a/Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
+++ b/Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
@@ -274,8 +274,14 @@ extension DefaultPublisher {
             Self.publisherStoppedCallback(handler: event.resultHandler)
             return
         }
-
-        routeProvider.changeRoutingProfile(to: routingProfile) { [weak self] result in
+        
+        guard let destination = activeTrackable?.destination else {
+            self.routingProfile = event.profile
+            Self.callback(value: Void(), handler: event.resultHandler)
+            return
+        }
+        
+        routeProvider.getRoute(to: destination.toCoreLocationCoordinate2d(), withRoutingProfile: event.profile) { [weak self] result in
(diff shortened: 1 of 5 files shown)


KacperKluka (approved):
In terms of functionality it LGTM but I haven't checked the code due to my poor Swift knowledge 😉

PR #476, opened 2022-11-23
Add ability to retry uploads in publisher example app

### LINKED ISSUE #475: Improve resilience of uploads in the publisher example app
We're going to be using the publisher example app for testing the SDK – Ablyans are going to get on bikes etc and travel around. It would be horrible if after all that effort, the upload of the location history data to S3 failed due to a dodgy Internet connection and their time was wasted. So I want to persist the list of uploads and add a retry button.

lawrence-forooghian:
**Note: this is based on #477, please review that one first.**

This persists all of the `LocationHistoryData` uploads that we try to perform, and adds a button for retrying a failed upload.

Here's a video of it in action (the code highlighted at the beginning is for simulating an error in the S3 upload):

https://user-images.githubusercontent.com/53756884/203615904-d28e6135-55a6-4363-92a1-44254f519e21.mov

Closes #475.

### DIFF
--- a/Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/Model/Uploads/UploadRequest.swift
+++ b/Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/Model/Uploads/UploadRequest.swift
@@ -1,8 +1,19 @@
 import Foundation
 import AblyAssetTrackingPublisher
 
-struct UploadRequest {
-    var data: LocationHistoryData
+struct UploadRequest: Codable {
+    enum UploadType: CustomStringConvertible, Codable {
+        case locationHistoryData(archiveVersion: String)
+        
+        var description: String {
+            switch self {
+            case .locationHistoryData:
+                return "Location history data"
+            }
+        }
+    }
+    
+    var type: UploadType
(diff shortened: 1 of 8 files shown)


Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/Model/Uploads/UploadRequest.swift
```diff
@@ -1,8 +1,19 @@
 import Foundation
 import AblyAssetTrackingPublisher
 
-struct UploadRequest {
-    var data: LocationHistoryData
+struct UploadRequest: Codable {
+    enum `Type`: CustomStringConvertible, Codable {
```
JakubJankowski, line 5: Not a biggest fan of using the metatype keyword as a name for the enum. Maybe we could name it `UploadType` or something along those lines instead?

lawrence-forooghian: Sure, changed 🙂

JakubJankowski (approved):
LGTM!

PR #477, opened 2022-11-23
Publisher example app: Make sure upload ends up in `failed` state after failure

lawrence-forooghian:
It’s currently falling through to the success case – a mistake in 19457a3.


JakubJankowski (approved):
Ahh, I missed that as well. Good catch.

PR #480, opened 2022-11-24
Emit and upload raw Mapbox location history data

### LINKED ISSUE #478: Emit raw Mapbox location recording data
Add an API like the Android SDK’s [`rawHistoryDataCallback`](https://github.com/ably/ably-asset-tracking-android/blob/393d87f7139ee5ecfad4d6a02fa9ef70a99f5f0b/publishing-sdk/src/main/java/com/ably/tracking/publisher/Publisher.kt#L232-L241).

### LINKED ISSUE #479: Upload raw Mapbox data to S3
After getting access to raw Mapbox location history data (see #478), upload this data to S3 like we currently do the `LocationHistoryData` objects.

lawrence-forooghian:
This closes #478 and closes #479, by adding an API to the publisher SDK to expose the raw Mapbox location history data, and then making the publisher example app upload this data to S3. See commit messages for more details.

Here's a demo of how it looks in the publisher example app:

https://user-images.githubusercontent.com/53756884/203791315-ce792a46-3577-49f7-a6d9-94d99d48d503.mov

### DIFF
--- a/Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/Model/Uploads/UploadsManager.swift
+++ b/Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/Model/Uploads/UploadsManager.swift
@@ -102,9 +102,46 @@ class UploadsManager: ObservableObject {
         performUpload(upload)
     }
     
+    @MainActor func uploadRawMapboxData(inTemporaryFile temporaryFile: TemporaryFile) {
+        guard s3Helper != nil else {
+            logger.info("Skipping upload of raw Mapbox data since S3 is not configured")
+            return
+        }
+        
+        let uploadId = UUID()
+        let dataFileURL: URL
+        do {
+            let storageDirectoryURL = try Storage.storageDirectoryURL(forUploadId: uploadId.uuidString)
+            try Storage.ensureStorageDirectoryExists(atURL: storageDirectoryURL)
+            dataFileURL = Storage.dataURL(forStorageDirectoryURL: storageDirectoryURL)
+            
+            try temporaryFile.stayAlive {
+                try FileManager.default.copyItem(at: temporaryFile.fileURL, to: dataFileURL)
+            }
+            logger.debug("Copied Mapbox raw history file to file: \(dataFileURL.path)")
+        } catch {
+            logger.error("Failed to copy Mapbox raw history file: \(error)")
+            return
+        }
+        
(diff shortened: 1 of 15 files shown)


Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/Model/Uploads/UploadsManager.swift
```diff
@@ -102,9 +102,46 @@ class UploadsManager: ObservableObject {
...
+        }
+        
+        let uploadId = UUID()
+        let dataFileURL: URL
+        do {
+            let storageDirectoryURL = try Storage.storageDirectoryURL(forUploadId: uploadId.uuidString)
+            try Storage.ensureStorageDirectoryExists(atURL: storageDirectoryURL)
+            dataFileURL = Storage.dataURL(forStorageDirectoryURL: storageDirectoryURL)
+            
+            try temporaryFile.stayAlive {
```
JakubJankowski, line 118: Could you elaborate on why you went with the "stayAlive" block approach?

Couldn't we just remove that file after the upload to S3 is successful? (I imagine that's when we want to clean it up?)

Right now it's kinda difficult for me to understand when the `temporaryFile` is expected to actually deallocate and remove the file underneath.

lawrence-forooghian: The problem is that it's the SDK that emits the URL of the file, but it's the example app that uploads to S3. So the SDK needs a way to know that the consumer of the file is finished with it, so that it can clear the file up. The responsibility for clearing up the file needs to live with the SDK, because there's no guarantee that the user of the SDK is even going to _use_ this file, so it can't be expected to clean it up.

lawrence-forooghian: > Right now it's kinda difficult for me to understand when the `temporaryFile` is expected to actually deallocate and remove the file underneath.

Do you think it's important for a user of the SDK to understand this? The point is that as long as they hold on to the `TemporaryFile` object, they have access to the file. And then at some point after they let go of it, the SDK will take care of clearing it up.

JakubJankowski (approved):

Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/Model/Uploads/UploadsManager.swift
JakubJankowski: Good point 👍

PR #481, opened 2022-12-01
Try to reduce time taken to get a trackable online

### LINKED ISSUE #425: Only the first trackable added to the publisher emits a `didChangeConnectionState:forTrackable:` callback
## Description

When the first trackable is added to a publisher, it emits a delegate `didChangeConnectionState: .online, forTrackable:` call. It does not do this for subsequent trackables added to the publisher (or sometimes, it does, but after a long delay).

Here's an example of how this looks in the publisher example app on my device (after the UI changes of #419):

https://user-images.githubusercontent.com/53756884/196199095-6473bf0c-fb3b-4144-a479-a00f18f5a1ce.mov

## Conditions to reproduce

I noticed this issue on my device running iOS 16.0 and on the iOS 15.5 simulator. It does not happen on the iOS 16.0 simulator.

lawrence-forooghian:
I noticed that on a stationary device, or on a simulator set to simulate a static location, the first trackable added would come online almost immediately, but subsequent ones would not come online or sometimes would come online after a long delay.

This is because a trackable will not come online until, after the trackable is added, we receive a location update from LocationService. The frequency of location updates from PassiveLocationManager is determined by that of CLLocationManager, which usually only emits location updates when the device has moved a sufficient distance (where “sufficient” is determined by the publisher’s resolution policy).

So, we provide a mechanism for the publisher to request a location update from CLLocationManager independently of the device’s motion, and we request a location update when any trackable (except for the first) is added.

It’s probably worth also explaining how Android behaves in this scenario and why. My notes here are based on commit 3a0d833.

There are two different things that happen in the Android asset tracking SDK, either of which has the side effect of causing the publisher to receive a location update when a new trackable is added:

1. The publisher calls DefaultMapbox.changeResolution, which in turn ends up removing and re-adding all of (Google SDK) FusedLocationProviderClient’s location observers (via its requestLocationUpdates method). The documentation for FusedLocationProviderClient.requestLocationUpdates implies that observers will always receive a location update after this method is called (although it might be an old location).

2. (Only in the case where the added trackable is set as the active trackable – i.e. when it is added using .track and not .add)

   The publisher calls DefaultMapbox.clearRoute, which calls (Mapbox SDK) MapboxNavigation.setNavigationRoutes. Through some process that I haven’t looked into, this ends up calling (Mapbox SDK) MapboxTripSession.kt’s navigatorObserver’s onStatus, which makes an explicit call to updateLocationMatcherResult, which emits a location update.

I think that the solution I’ve implemented for iOS turns Android side effect 1 into an explicit behaviour.

Closes #425.


Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
```diff
@@ -71,6 +71,16 @@ class DefaultLocationService: LocationService {
...
     }
     
+    func requestLocationUpdate() {
+        /*
+         From reading the ``CLLocationManager/startUpdatingLocation`` documentation, this seems to be the only programmatic way to provoke it into emitting a location event:
+        
+        > Calling this method several times in succession does not automatically result in new events being generated. Calling stopUpdatingLocation() in between, however, does cause a new initial event to be sent the next time you call this method.
+        */
+        locationManager.systemLocationManager.stopUpdatingLocation()
+        locationManager.systemLocationManager.startUpdatingLocation()
```
KacperKluka, line 81: I wonder how this affects the Mapbox SDK 🤔 
1. Have you checked whether this doesn't break things like location history recording? 
2. Will there be a big delay in location updates between the "stop" and "start" operations?
3. Additionally, do we know what happens when you stop and start location updates? Will it consume a lot of resources to do that?

KacperKluka (commented):
Overall I like the idea and it's great that you've spotted and fixed it 🙇 I just have a few questions about the stopping and starting of location updates 😉

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
lawrence-forooghian: These are all good questions, I'll look into it and get back to you 🙂

ikbalkaya (commented):
The code looks good to me. Like Kacper I also would be interested to know whether stopping / starting location service might have side effects on other areas

Sources/AblyAssetTrackingPublisher/Services/DefaultLocationService.swift
lawrence-forooghian: I've added [a script](https://github.com/ably/ably-asset-tracking-swift/tree/425-only-first-trackable-added-comes-online-when-stationary/Tools/Analysis/AnalyzeLocationEventFrequencies) for analysing the logs from the publisher example app, to find out if the frequencies of location updates and recorded locations is affected. My brief investigation using the iOS simulator suggests that they are not. However, I want to test on a device too, and for that we'll probably need to add the ability to log to a file. Once I've done that and gathered some results from a device, I'll post my findings here in detail.

As for resources (which I'm mainly taking to mean battery life), I do not know what the impact would be. I could have a think about how to investigate it, but I don't have any ideas off the top of my head.

lawrence-forooghian:
Thanks for the review, @ikbalkaya. I've just posted a reply to Kacper's questions, about the current state of my findings: https://github.com/ably/ably-asset-tracking-swift/pull/481#discussion_r1052140016

PR #484, opened 2022-12-05
Some refactors around calling of completion handlers in `DefaultPublisher`

lawrence-forooghian:
A few refactors that I made in preparation for #341. Please review commit by commit, and see commit messages for more details.

Issue #357, opened 2022-08-25
Rethink whether it's a good idea to return success if the 'same' trackable is being added.


ikbalkaya:
We return true or callback with success when an asset with the same trackable id of an existing one is being added. This has following potential problems

* This is a bit misleading for clients, Instead of returning true, we could simply callback with a 'state' or an 'error'. I do not think there is a need to throw an exception here.
* Equality of trackables are based on trackable ids. I think we should rethink what it means for two trackables to be 'equal', for example considering gps accuracy etc. 

See  https://github.com/ably/ably-asset-tracking-swift/issues/340#issuecomment-1225802013_

deanna-lad:
The corresponding Jira issue is https://ably.atlassian.net/browse/SDK-2605

PR #485, opened 2022-12-05
Do not repeat work when we try to add a trackable twice

### LINKED ISSUE #341: Handle a scenario where the same trackable is being added multiple times

lawrence-forooghian:
**Please note that this PR is based on top of #484.**

This handles the case where a user adds a trackable and then tries to add another trackable with the same ID before the first add operation has completed. It ensures that we do not repeat the work of joining Ably presence, instead saving the completion handler for the second add operation and calling it when the first add operation completes.

This is copied from the Android implementation (as of 3a0d833).

It’s worth pointing out that this continues our existing notion of trackables being “the same” when adding them – that is, that we only compare them by ID. This might not be ideal and we’ve already opened issue #357 to revisit this.

Closes #341.


Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
```diff
@@ -304,6 +307,12 @@ extension DefaultPublisher {
     }
 
     private func performAddOrTrack(_ trackable: Trackable, completion: Callback<Void>){
+        guard !duplicateTrackableGuard.isCurrentlyAddingTrackableWithId(trackable.id) else {
+            self.logHandler?.debug(message: "\(Self.self): Trackable with ID \(trackable.id) is already being added; saving completion handler.", error: nil)
+            duplicateTrackableGuard.saveDuplicateAddCompletionHandler(completion, forTrackableWithId: trackable.id)
```
maratal, line 312: I know that's because I'm not familiar with this, but I can't find where `saveDuplicateAddCompletionHandler` is called to add the very first callback?

Tests/PublisherTests/DefaultPublisher/DefaultPublisherTests.swift
```diff
@@ -353,6 +353,92 @@ class DefaultPublisherTests: XCTestCase {
         wait(for: [expectation], timeout: 5.0)
     }
     
+    func testAdd_whenTrackableWithSameIdIsCurrentlyBeingAdded_itWaitsForTheFirstAddToComplete_andWhenTheFirstAddSucceeds_theSecondAddSucceedsToo() {
+        var numberOfTimesConnectCalled: Int32 = 0
+        var connectCompletion: ResultHandler<Void>? = nil
+        let connectCalledExpectation = expectation(description: "ablyPublisher’s connect method is called")
+        ablyPublisher.connectCompletionHandler = { completion in
+            OSAtomicIncrement32(&numberOfTimesConnectCalled)
```
JakubJankowski, line 361: Hmm, seems like `OSAtomicIncrement32` has been deprecated since iOS 10:
```
@available(iOS, deprecated: 10.0, message: "Use atomic_fetch_add_explicit(memory_order_relaxed) from <stdatomic.h> instead")
``` 
I'm not a fan of using code that's been deprecated, especially for so long. 
Seems like using the suggested `atomic_fetch_add_explicit(memory_order_relaxed)` would be a real pain though.

There's however an Apple developed atomics package for Swift, maybe we could switch to it instead?
https://github.com/apple/swift-atomics

Adding a new dependency just to use atomics in two tests seems like an overkill though, so I'll leave it up to you.

JakubJankowski (approved):
LGTM 👍

Tests/PublisherTests/DefaultPublisher/DefaultPublisherTests.swift
lawrence-forooghian: Yeah, I'd prefer to not add another package unless we need it in more than 2 places. I've just switched it to use a serial queue for synchronisation instead.

Sources/AblyAssetTrackingPublisher/DefaultPublisher.swift
lawrence-forooghian: We don't use the `DuplicateTrackableGuard` to save the callback for the _first_ `track` call for the trackable. It's only used for saving the completion handler for the subsequent calls (hence the `duplicate` in `saveDuplicateAddCompletionHandler`).

maratal (approved):
LGTM

Issue #494, opened 2022-12-07
Make logging consistent throughout codebase


lawrence-forooghian:
We want it to be easy to emit log messages (as opposed to a copy and paste from somewhere else in the codebase) and we want to be able to easily parse log messages (hence have a consistent structure).

- The ability to specify a system / subsystem
- Display line number and file name
- Making sure that Ably always logs through our logger (see https://github.com/ably/ably-cocoa/issues/1554)

PR #508, opened 2022-12-12
Add info to log messages about which subsystem emitted message

lawrence-forooghian:
This adds a tag like `[assetTracking.subsystem1.subsystem2]` to each log message, identifying which subsystem (which might be a type, or might be the ably-cocoa SDK) emitted the message.

For example, a log from ably-cocoa inside the Publisher SDK would be tagged `[assetTracking.publisher.DefaultAbly.ablySDK]`. (This might turn out to be a little too verbose, but we can tweak that later.)

This allows us to always identify the output of the ably-cocoa SDK, and also allows us to remove all of the ad-hoc logging of `Self.self`.

Relates to #494.

### DIFF
--- a/Sources/AblyAssetTrackingInternal/Logging/DefaultHierarchicalLogHandler.swift
+++ b/Sources/AblyAssetTrackingInternal/Logging/DefaultHierarchicalLogHandler.swift
@@ -1,0 +1,34 @@
+import AblyAssetTrackingCore
+
+/// Provides an implementation of ``HierarchicalLogHandler`` by wrapping another instance of ``LogHandler``. It will insert a string such as "[asset-tracking.publisher.DefaultPublisher]" into the log messages, representing its list of subsystems.
+public struct DefaultHierarchicalLogHandler: HierarchicalLogHandler {
+    private var logHandler: LogHandler
+    private var subsystemNames: [String] // Lowest granularity first
+    
+    /// Creates an instance of ``DefaultHierarchicalLogHandler`` that writes messages to a given ``LogHandler``. The lowest-granularity subsystem of the returned log handler will always be "assetTracking".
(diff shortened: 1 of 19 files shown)


Sources/AblyAssetTrackingInternal/Logging/DefaultHierarchicalLogHandler.swift
```diff
@@ -0,0 +1,30 @@
...
+    /// Creates an instance of ``DefaultHierarchicalLogHandler`` that writes messages to a given ``LogHandler``. The lowest-granularity subsystem of the returned log handler will always be "assetTracking".
+    /// - Parameters:
+    ///   - logHandler: A log handler to write messages to. If this is nil, the initializer will return nil.
+    ///   - subsystem: A subsystem to add to the list
+    public init?(logHandler: LogHandler?, subsystem: Subsystem? = nil) {
+        guard let logHandler = logHandler else {
+            return nil
+        }
+        self.logHandler = logHandler
+        self.subsystemNames = ["assetTracking"] + [subsystem?.name].compactMap { $0 }
```
JakubJankowski, line 17: Nice one-liner approach to unwrapping the `subsystem?.name` optional value. I'm not sure if it's the most readable though. Wouldn't `if let name = subsystem?.name` be easier to read?

Up to you though, I don't really feel strongly either way.

JakubJankowski (approved):
LGTM!

Sources/AblyAssetTrackingInternal/Logging/DefaultHierarchicalLogHandler.swift
lawrence-forooghian: Fair point! I've changed it to your suggestion.

PR #509, opened 2022-12-12
Add file name and line number info to log messages

lawrence-forooghian:
This adds a tag like `@(MyFile.swift:130)` to each log message, identifying the source code file and line number from which the log message was emitted.

Please review commit-by-commit (it has a couple of refactors that might be difficult to review otherwise) and see commit messages for more details.

Closes #494.

PR #511, opened 2022-12-12
Create a library for parsing the logs from the SDK and example apps

### LINKED ISSUE #510: Create a package for parsing logs
This will be useful for any tasks in which we want to analyse logs emitted by the library, for example answering [this question](https://github.com/ably/ably-asset-tracking-swift/pull/481#discussion_r1037181957).

I want to create a library that will allow us to build scripts/tooling in Swift for analysing the logs emitted by the library and the example apps. It should parse a log file and turn it into an array of structs representing a line (with information like timestamp, log level, etc).

lawrence-forooghian:
**Note: This PR is based on top of #509. Please review that one first.**

This adds a new `LogParser` library in the `Tools` directory, which is able to parse the logs from the SDK and its example apps.

It’s intended to make it easy to write quick scripts for analysing the log output of the SDK.

See commit messages for more details.

Closes #510.


Tools/LogParser/README.md
```diff
@@ -0,0 +1,9 @@
+# LogParser
+
+A package for parsing the log output of the Ably Asset Tracking SDK and its example apps.
```
QuintinWillison, line 3: Might be worth being more specific here, as I'm assuming this isn't compatible with Android SDK logs:

```suggestion
A package for parsing the log output of the Ably Asset Tracking Swift SDKs and their example apps.
```

QuintinWillison (approved):
My language and domain knowledge isn't great so I definitely don't fully understand everything I'm looking at, however I do think this is an interesting addition and I'll be interested to see how it ends up being used.

davyskiba (approved):

Tools/LogParser/README.md
lawrence-forooghian: Fair point, updated it 👍

PR #512, opened 2022-12-12
Align logging code in example apps

lawrence-forooghian:
This just copies the logging code from the publisher example app to the subscriber one. Means that we get the same format and hence will be able to use the same code for parsing their output in #510.


Examples/SubscriberExample/Sources/Screens/Logging/SubscriberLogger.swift
```diff
@@ -3,10 +3,10 @@ import AblyAssetTrackingCore
 import Logging
 
 class SubscriberLogger: AblyAssetTrackingCore.LogHandler {
-    let swiftLog: Logger
+    private let swiftLog: Logger
```
maratal, line 6: Is there a reason why this variable name takes the name of the language into its own name?

maratal: Oh I see, that's from https://github.com/apple/swift-log, okay than)

maratal (approved):
LGTM, but maybe `swiftLog` var still can be cleaned from carrying framework name in it imho.

lawrence-forooghian:
Thanks @maratal for the suggestion, I've made that change: fa81352874506ec972788df45bd09caf194c37f5

PR #514, opened 2022-12-14
Use Fastlane’s `trainer` to generate JUnit report

### LINKED ISSUE #513: Failing tests are sometimes missing from JUnit report uploaded to test observability server
For example, see https://github.com/ably/ably-asset-tracking-swift/actions/runs/3679080639/jobs/6235577712; the following tests are failing (and many others):

- `ChannelModesTests.testShouldCreateOnlyOnePublisherAndOneSubscriberConnection()`
- `PublisherAndSubscriberSystemTests.testSubscriberReceivesAssetConnectionStatusWithBicycleProfile()`
- `PublisherAndSubscriberSystemTests.testSubscriberReceivesAssetConnectionStatusWithCarProfile()`
- `PublisherAndSubscriberSystemTests.testSubscriberReceivesPublisherMessageWithBicycleProfile()`
- `PublisherAndSubscriberSystemTests.testSubscriberReceivesPublisherMessageWithCarProfile()`

But none of those appear in [the JUnit report uploaded to the test observability server](https://test-observability.herokuapp.com/repos/ably/ably-asset-tracking-swift/uploads/88f56331-6f0c-413e-98e4-50277338216a/junit_report_xml).

Also, the number of tests per upload seems to be varying wildly (see the "Total number of tests" column [here](https://test-observability.herokuapp.com/repos/ably/ably-asset-tracking-swift/uploads)).

lawrence-forooghian:
It seems that `xcpretty`’s JUnit report generation doesn’t work. For example, when we give it the `xcodebuild` output of a test run which had lots of crashes (e.g. https://github.com/ably/ably-asset-tracking-swift/issues/513#issuecomment-1349970753), it completely ignores those tests and doesn’t include them in its results.

So I’m switching to using `trainer`, which we already use in ably-cocoa. It directly reads the `.xcresult` bundle emitted by Xcode and gives much better results (for example, in https://github.com/ably/ably-asset-tracking-swift/issues/513#issuecomment-1349970753 it correctly identifies the crashed tests as failures).

Closes #513.


Scripts/test.sh
```diff
@@ -23,18 +23,38 @@ struct Secrets {
...
-set -o pipefail && xcodebuild test -scheme "ably-asset-tracking-swift-Package" -destination 'platform=iOS Simulator,name=iPhone 12' \
-	| xcpretty --report junit
+set -o pipefail && xcodebuild test -scheme "ably-asset-tracking-swift-Package" -destination 'platform=iOS Simulator,name=iPhone 12' -derivedDataPath "${derived_data_dir}" \
+	| xcpretty
 xcodebuild_exit_status=$?
 set -e
 
-# test-observability-action looks for .junit files
-cp build/reports/junit.xml results.junit
+# Create a directory to store all of the files we’ll pass to actions
```
QuintinWillison, line 44: I'm not sure "pass to actions" is the best turn of phrase, especially for those with less experience with GitHub's ecosystem. Perhaps:

```suggestion
# Create a directory to store all of the files we wish to persist outside of the GitHub
# workflow that called this script. That workflow uses the upload-artifact action to
# archive and save the contents of this folder for later inspection.
```

QuintinWillison (approved):
Nice. This approach looks really clean and slick. Just one thought around commentary...

Scripts/test.sh
lawrence-forooghian: Good point. I've replaced it with the following:

```
# Create a directory to store all of the files that we wish to pass to
# subsequent actions in the GitHub workflow that called this script.
```

I didn't want to get too verbose, since it isn't just upload-artifact that we use – there's also test-observability-action.

I think this still keeps to the spirit of your suggested change, though?

QuintinWillison: Nice. Thanks!

PR #516, opened 2022-12-15
LogParser: Properly extract the message and error description from example app logs

lawrence-forooghian:
This removes this caveat from `ExampleAppSDKLogLine#message`:

> The ``SDKLogMessage/message`` of this value is not necessarily the exact value that was passed as the `message` argument of AblyAssetTrackingCore’s `LogHandler.logMessage(level:message:error:)`.  It may contain a suffix added by the example apps’ log handlers, to add a trailing full stop and information about the `error` argument.

Now, `#message` contains the exact value passed as the `message` argument to `LogHandler.logMessage(level:message:error:)`, and `#errorDescription` contains the `localizedDescription` of the error passed as the `error` argument to that method.

I want access to the exact log message emitted by the SDK, without any additional characters, so that I can for example emit some JSON in a log message from the SDK and then parse this JSON in an app that reads the logs.

### DIFF
--- a/Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/Logger/PublisherLogger.swift
+++ b/Examples/PublisherExampleSwiftUI/PublisherExampleSwiftUI/Logger/PublisherLogger.swift
@@ -11,17 +11,18 @@ class PublisherLogger: AblyAssetTrackingCore.LogHandler {
     
     func logMessage(level: LogLevel, message: String, error: Error?) {
         let errorString = error?.localizedDescription
-        switch level {
-        case .verbose:
-            logger.log(level: .trace, "\(message). \(errorString ?? "")")
-        case .info:
-            logger.log(level: .info, "\(message). \(errorString ?? "")")
-        case .debug:
-            logger.log(level: .debug, "\(message). \(errorString ?? "")")
-        case .warn:
-            logger.log(level: .warning, "\(message). \(errorString ?? "")")
-        case .error:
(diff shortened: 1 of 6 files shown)


Tools/LogParser/Examples/CommandLineExample/example.txt
```diff
@@ -1,3 +1,4 @@
-2022-12-13T09:06:06.341000-03:00 debug: [assetTracking.someComponent] Here is a message
+2022-12-13T09:06:06.341000-03:00 debug: [noError] [assetTracking.someComponent] Here is a message
 Some other line
-2022-12-13T09:06:10.729000-03:00 info: [assetTracking.someOtherComponent]@(MyFile.swift:130) Here is another message
+2022-12-13T09:06:10.729000-03:00 info: [noError] [assetTracking.someOtherComponent]@(MyFile.swift:130) Here is another message
+2022-12-13T09:06:32.282000-03:00 info: [error(len:24): Here is an error message] [assetTracking.someComponent] Here is a message with an attached error
```
maratal, line 4: Isn't it easy to detect end of message by looking for the first "]" symbol?

lawrence-forooghian: What if the message contains a `]`?

maratal: Fair point, for the sake of simplicity length makes sense, otherwise parser complexity grows quickly.

maratal (approved):

PR #517, opened 2022-12-15
Publisher example: Add a setting for logging `LocationHistoryData` JSON

lawrence-forooghian:
While testing a fix for #425, I wanted to have access to the location history data in a log message.

This adds a switch in a new "Developer settings" section of the Settings page, which when turned on will emit a JSON serialization of the location history data when it’s received from the SDK. It’s turned off by default, because it's only for my specific use case.


ikbalkaya (approved):
LGTM

maratal (approved):

PR #598, opened 2023-03-02
Fix early-exit bug in `ConnectionConfiguration.getClientOptions`

lawrence-forooghian:
For some reason, this method was exiting early when an `authCallback` was specified. This definitely seems like an accident, and I can’t find anything to suggest any motivation for it — all of the skipped code seems like it should be run even in the presence of an `authCallback`.


AndyTWF (approved):
Good catch! LGTM :)

PR #605, opened 2023-03-02
Add ability to internally specify custom Ably host

lawrence-forooghian:
We need this for the upcoming `NetworkConnectivityTests`, so that we can connect to the local SDK test proxy.

### DIFF
--- a/Sources/AblyAssetTrackingInternal/AblyWrapper/SDK/AblyCocoaSDK.swift
+++ b/Sources/AblyAssetTrackingInternal/AblyWrapper/SDK/AblyCocoaSDK.swift
@@ -125,8 +125,14 @@ struct AblyCocoaSDKEventListener: AblySDKEventListener {
 public class AblyCocoaSDKRealtimeFactory: AblySDKRealtimeFactory {
     public init() {}
     
-    public func create(withConfiguration configuration: ConnectionConfiguration, logHandler: InternalARTLogHandler) -> AblySDKRealtime {
-        let realtime = ARTRealtime(options: configuration.getClientOptions(logHandler: logHandler, remainPresentForMilliseconds: configuration.remainPresentForMilliseconds))
+    public func create(withConfiguration configuration: ConnectionConfiguration, logHandler: InternalARTLogHandler, host: Host?) -> AblySDKRealtime {
+        let clientOptions = configuration.getClientOptions(
+            logHandler: logHandler,
+            remainPresentForMilliseconds: configuration.remainPresentForMilliseconds,
+            host: host
+        )
+
+        let realtime = ARTRealtime(options: clientOptions)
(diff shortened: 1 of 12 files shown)


Sources/AblyAssetTrackingInternal/AblyWrapper/SDK/AblyCocoaSDK.swift
```diff
@@ -125,8 +125,14 @@ struct AblyCocoaSDKEventListener: AblySDKEventListener {
 public class AblyCocoaSDKRealtimeFactory: AblySDKRealtimeFactory {
     public init() {}
     
-    public func create(withConfiguration configuration: ConnectionConfiguration, logHandler: InternalARTLogHandler) -> AblySDKRealtime {
-        let realtime = ARTRealtime(options: configuration.getClientOptions(logHandler: logHandler, remainPresentForMilliseconds: configuration.remainPresentForMilliseconds))
+    public func create(withConfiguration configuration: ConnectionConfiguration, host: Host?, logHandler: InternalARTLogHandler) -> AblySDKRealtime {
```
AndyTWF, line 128: Most languages would have the `host` parameter last, as it's optional. Should we do the same here? :)

Sources/AblyAssetTrackingInternal/Configuration/Host.swift
```diff
@@ -0,0 +1,12 @@
+/// Internal-only type for configuring some properties of `ARTClientOptions` which we do not wish to expose through the public `ConnectionConfiguration` interface. Intended for use in testing.
+public struct Host {
```
AndyTWF, line 2: If this is internal only, should its accessibility modifier be internal (or omitted as that's the default)?

lawrence-forooghian: If it were `internal`, that would mean that it was only accessible inside the `AssetTrackingInternal` library – but we need it to be visible to callers of the `public` method `create(configuration:,host:,logHandler:)`.

It's a bit confusing — the `AssetTrackingInternal` library is intended only to be used internally by the SDK, but since the rest of the SDK lives in other libraries, all of the functionality it wishes to expose to the rest of the SDK needs to be public.

Sources/AblyAssetTrackingInternal/AblyWrapper/SDK/AblyCocoaSDK.swift
lawrence-forooghian: Done 🙂

Sources/AblyAssetTrackingInternal/Configuration/Host.swift
AndyTWF: Ah yep, gotcha!

AndyTWF (approved):

PR #611, opened 2023-03-08
Update Swift version in `Package.swift` to 5.7

lawrence-forooghian:
And address the deprecation warning this introduces.

This should have been done alongside c918aea.

I believe that this compiler version change introduces a new version (2) of `Package.resolved`, but calling `swift package resolve` alone doesn’t force the compiler to re-generate this file – seems like it will only do so next time we have an actual change in our resolved dependencies. So I’ll let it perform that migration when it needs to in the future.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.7
 // The swift-tools-version declares the minimum version of Swift required to build this package.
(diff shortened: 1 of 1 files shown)


lawrence-forooghian:
Putting back into draft since struggling to get CI green.

AndyTWF (approved):

PR #612, opened 2023-03-09
Implement subscriber network connectivity tests

### LINKED ISSUE #539: Implement the same subscriber `NetworkConnectivityTests` as Android
We’ll do this after https://github.com/ably/ably-asset-tracking-swift/issues/538. Depends on https://github.com/ably/ably-asset-tracking-android/pull/924.

lawrence-forooghian:
Resolves #539. See commit messages for more details.


Tests/SystemTests/NetworkConnectivityTests/Subscriber/Helper/SubscriberMonitor.swift
```diff
@@ -0,0 +1,382 @@
...
+            guard let expectedPublisherResolution else {
+                logHandler.debug(message: "(SKIP) expectedPublisherResolution = nil", error: nil)
+                return
+            }
+
+            logHandler.debug(message: "(WAITING) expectedPublisherResolution = \(expectedPublisherResolution)", error: nil)
+
+            let lastPublisherResolution = try listenForExpectedPublisherResolution(waiterContext: waiterContext)
+
+            // TODO this seems a bit redundant – listenForExpectedPublisherResolution will either throw or return expectedPublisherResolution, won't it?
```
lawrence-forooghian, line 242: @AndyTWF I was wondering if you had any thoughts on this — did I miss something?

Tests/SystemTests/NetworkConnectivityTests/Subscriber/Helper/SubscriberMonitor.swift
```diff
@@ -0,0 +1,382 @@
...
+                // no expected location set - skip assertion
+                logHandler.debug(message: "(SKIP) expectedLocationUpdate = null", error: nil)
+                return
+            }
+
+            logHandler.debug(message: "(WAITING) expectedLocationUpdate = \(expectedLocation)", error: nil)
+
+            let lastPublishedLocation = try listenForExpectedLocationUpdate(waiterContext: waiterContext)
+
+            // TODO this seems a bit redundant – listenForExpectedLocationUpdate will either throw or return expectedLocation, won't it?
```
lawrence-forooghian, line 335: @AndyTWF I was wondering if you had any thoughts on this — did I miss something?

Tests/SystemTests/NetworkConnectivityTests/Subscriber/Helper/SubscriberMonitor.swift
```diff
@@ -0,0 +1,382 @@
...
+                logHandler.debug(message: "(SKIPPED) expectedSubscriberResolution = nil", error: nil)
+                return
+            }
+
+            logHandler.debug(message: "(WAITING) preferredSubscriberResolution = \(expectedSubscriberResolution)", error: nil)
+
+            let publisher = subscriberResolutionPreferences.filter { $0 == expectedSubscriberResolution }
+            let preferredSubscriberResolution = try first(publisher, label: "Wait for subscriber resolution preference \(expectedSubscriberResolution)", waiterContext: waiterContext)
+
+            // TODO this seems a bit redundant – the call to `first` will either throw or return expectedSubscriberResolution, won't it?
```
lawrence-forooghian, line 373: @AndyTWF I was wondering if you had any thoughts on this — did I miss something?

Tests/SystemTests/NetworkConnectivityTests/Subscriber/SubscriberNetworkConnectivityTests.swift
```diff
@@ -0,0 +1,384 @@
...
+
+            }
+        }
+
+        // Restart the fault to simulate the publisher going away whilst we're offline
+
+        try subscriberMonitorFactory.forActiveFault(
+            label: "[fault active] publisher shutdown for disconnect test",
+            locationUpdate: thirdLocationUpdate,
+            publisherResolution: /* thirdPublisherResolution, */ nil, /* TODO I’m not convinced that thirdPublisherResolution is right. Since shutdownAblyPublishing leaves presence with a hardcoded presence data, and hence a hardcoded resolution (accuracy: .balanced, desiredInterval: 1, minimumDisplacement: 0) that’s the only resolution the subscriber can expect to definitely receive */
```
lawrence-forooghian, line 356: @AndyTWF I was wondering if you had any thoughts on this — did I miss something?

Tests/SystemTests/NetworkConnectivityTests/Subscriber/Helper/SubscriberMonitor.swift
AndyTWF: Yeah - it looks like first() will throw if it times out waiting for a value, so I think you can remove this check.

Tests/SystemTests/NetworkConnectivityTests/Subscriber/Helper/SubscriberMonitor.swift
AndyTWF: As per above :)

Tests/SystemTests/NetworkConnectivityTests/Subscriber/Helper/SubscriberMonitor.swift
AndyTWF: As per above :)

Tests/SystemTests/NetworkConnectivityTests/Subscriber/SubscriberNetworkConnectivityTests.swift
AndyTWF: I believe in Android it was because the last thing on the StateFlow was the third resolution, so checking it again would result in the same value being emitted to a new collector?

Tests/SystemTests/NetworkConnectivityTests/Subscriber/Helper/SubscriberMonitor.swift
lawrence-forooghian: Cool, will remove 👍

Tests/SystemTests/NetworkConnectivityTests/Subscriber/Helper/SubscriberMonitorFactory.swift
```diff
@@ -0,0 +1,319 @@
...
+            case .nonfatal:
+                expectedState = .online
+            case .nonfatalWhenResolved:
+                expectedState = .offline
+            }
+
+            let failureStates: Set<ConnectionState>
+            switch faultType {
+            case .fatal:
+                failureStates = [.offline]
```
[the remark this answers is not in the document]
lawrence-forooghian: Seems like it would be. It's a bit of a moot point right now since there are no fatal tests anyway. Like I said, I just copied the assertions from Android ([here's this one](https://github.com/ably/ably-asset-tracking-android/blob/f6d163fc47aa329e6f7e41d5cb4a15ee87c4bc43/subscribing-sdk/src/androidTest/java/com/ably/tracking/subscriber/NetworkConnectivityTests.kt#L645)) without applying much of a critical eye. Reckon I should change it here or keep it consistent with Android?

Tests/SystemTests/NetworkConnectivityTests/Subscriber/SubscriberNetworkConnectivityTests.swift
lawrence-forooghian: But `testResources.shutdownAblyPublishingAsync` causes the publisher to emit a resolution update, doesn't it? Wouldn't we expect that to be received by the subscriber?

AndyTWF: So yes, but it never reaches the subscribers stateflow.

As the presence event is a `LEAVE`, the subscriber doesn't actually process it as a resolution update, it just marks the publisher as gone (see `updateForPresenceMessagesAndThenEmitStateEventsIfRequired` in Android).

So I would expect the "last" emitted value to be that of the last presence `update` - which in this case is the `thirdPublisherResolution`.

Tests/SystemTests/NetworkConnectivityTests/Subscriber/Helper/SubscriberMonitorFactory.swift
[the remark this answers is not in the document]
AndyTWF: I think its adding here and then creating an issue on the Android side to bring parity

Tests/SystemTests/NetworkConnectivityTests/Subscriber/SubscriberNetworkConnectivityTests.swift
lawrence-forooghian: Cool, thanks for the explanation. I've opened https://github.com/ably/ably-asset-tracking-swift/pull/633 which ignores `ABSENT` and `LEAVE` actions containing resolutions, and that makes the original assertion work.

AndyTWF (approved):

PR #615, opened 2023-03-13
Don’t use Sourcery log handler mock where concurrent access is possible

lawrence-forooghian:
We seem to have had an increased number of crashes in the test suite recently, e.g. [this one](https://test-observability.herokuapp.com/repos/ably/ably-asset-tracking-swift/failures/3029abb9-8709-4a41-b9d5-425c35a011a7). I briefly investigated these and I _think_ it’s due to concurrent access of mutable shared state in `InternalLogHandlerMock`.

So, don’t use the Sourcery mock in these situations, and document why.

I suspect that I introduced these crashes in 0e8515c, which adds logging to the `Callback` struct. Its messages are logged from multiple threads.


JakubJankowski (approved):
Nice find, LGTM 👍

PR #617, opened 2023-03-15
Feature/593 subscriber properties

### LINKED ISSUE #593: Subscriber Worker Queue: Subscriber Properties implementation
Implement the SubscriberProperties from the Android AAT in Swift. The class definition itself is being added in https://github.com/ably/ably-asset-tracking-swift/pull/573, we just need to add the members.

At this point, we're proposing that we don't switch things out in the "old" subscriber - but migrate things along as we introduce workers.

JakubJankowski:
This PR introduces the implementation of SubscriberProperties, to be used by WorkerQueue on the Subscriber side of the sdk.
I tried to keep the implementation as close to Android's approach as possible. However we've decided to, at least for now, stick to the Delegation pattern to notify the API consumers about state changes, instead of switching to the Kotlin's streams (SharedFlow) approach.

Resolves #593


lawrence-forooghian:
@JakubJankowski did you see the CI failure here? I’ll put it back in to draft for now.

JakubJankowski:
As mentioned in one of our calls - any suggestions on how to handle the naming of `LocationsPublishingState`, `TrackableState` and `ConnectionState` are welcome.
I've decided for now to preserve the `ConnectionState` name, so that the API wouldn't change too much, but the `TrackableState` would fit better and would align with the naming in Kotlin.

Sources/AblyAssetTrackingSubscriber/WorkerQueue/SubscriberWorkerQueueProperties.swift
```diff
@@ -1,7 +1,181 @@
 import Foundation
 import AblyAssetTrackingInternal
 
-struct SubscriberWorkerQueueProperties: WorkerQueueProperties
-{
+struct SubscriberWorkerQueueProperties: WorkerQueueProperties {
     public var isStopped = false
+    var presenceData: PresenceData
+    let ablySubscriber: Subscriber
```
JakubJankowski, line 7: I don't like this bit tbh - the way I implemented it is we pass the ablySubscriber instance here so that we can set it as a sender when firing delegate methods, for example:
```
delegate?.subscriber(sender: ablySubscriber, didUpdatePublisherPresence: publisherPresence)
```
The main reason for that was to avoid changing the api by changing the `sender` type, but it complicates the rest of the code quite a bit, having to introduce this weird `configureWorkerQueue` method on default subscriber... So maybe in this case changing the api a bit would be justified, and the `sender` in delegate methods should just be of type `SubscriberWorkerQueueProperties `. Thoughts?

Sources/AblyAssetTrackingSubscriber/WorkerQueue/SubscriberWorkerQueueProperties.swift
```diff
@@ -1,7 +1,181 @@
...
+    
+    private var updatingResolutions: [String: [Resolution?]] = [:]
```
lawrence-forooghian, line 9: It would be helpful to know which version of the Android codebase you have based your work on. `updatingResolutions` doesn't exist in `v1.6.3`, which is what we're basing our current work on.

Sources/AblyAssetTrackingCore/Model/ConnectionState.swift
```diff
@@ -8,7 +8,14 @@ public enum ConnectionState: Int {
...
      */
     case online
-
+    
+    /**
+     Asset state is publishing when its locations are being published but it is not able to detect subscribers or receive data from them.
+     This state allows the asset to be actively tracked, however, its features are limited compared to the online state.
+     This state can change to either online, offline or failed.
+     */
+    case publishing
```
lawrence-forooghian, line 17: I was wondering why you'd added this as part of this PR — does it need to be here? I've created https://github.com/ably/ably-asset-tracking-swift/issues/625 for more broadly bringing our `TrackableState` in line with that of Android, perhaps we can do it as part of that instead?

Sources/AblyAssetTrackingPublisher/State/LocationsPublishingState.swift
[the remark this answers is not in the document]
lawrence-forooghian: This change is now contained as part of https://github.com/ably/ably-asset-tracking-swift/pull/626.

lawrence-forooghian (commented):
> As mentioned in one of our calls - any suggestions on how to handle the naming of `LocationsPublishingState`, `TrackableState` and `ConnectionState` are welcome.

It's not clear to me that this needs to be addressed within the scope of this PR — what do you think? I’ve implemented some of these changes in #626, though, and also created issue #625.

Sources/AblyAssetTrackingSubscriber/WorkerQueue/SubscriberWorkerQueueProperties.swift
JakubJankowski: Yeah, that's my mistake - used one of the newest commits as reference, forgot we're supposed to stick with 1.6.3. I'll rewrite it to the 1.6.3 state.

Sources/AblyAssetTrackingSubscriber/WorkerQueue/SubscriberWorkerQueueProperties.swift
```diff
@@ -3,4 +3,185 @@ import Foundation
...
+    }
+    
+    mutating func notifyDidFailWithError(error: ErrorInformation) {
+        guard let subscriber = subscriber
+        else { return }
+        delegate?.subscriber(sender: subscriber, didFailWithError: error)
+    }
+    
+    mutating func notifyResolutionsChanged(resolutions: [Resolution]) {
+        if !resolutions.isEmpty {
```
lawrence-forooghian, line 162: Can we use `guard`?

Sources/AblyAssetTrackingSubscriber/WorkerQueue/SubscriberWorkerQueueProperties.swift
```diff
@@ -3,4 +3,185 @@ import Foundation
...
+    mutating func updateForChannelConnectionStateChangeAndThenDelegateStateEventsIfRequired(stateChange: ConnectionStateChange) {
+        lastChannelConnectionStateChange = stateChange
+        delegateStateEventsIfRequired()
+    }
+    
+    mutating func updateForPresenceMessagesAndThenDelegateStateEventsIfRequired(presenceMessages: [Presence]) {
+        for presenceMessage in presenceMessages {
+            if (presenceMessage.data.type == .publisher) {
+                
+                if (presenceMessage.action == .leave || presenceMessage.action == .absent) {
```
lawrence-forooghian, line 68: Can we use a `switch`?

Sources/AblyAssetTrackingSubscriber/WorkerQueue/SubscriberWorkerQueueProperties.swift
```diff
@@ -1,7 +1,181 @@
...
+            lastEmittedValueOfIsPublisherVisible = isPublisherVisible
+            notifyPublisherPresenceUpdated(isPublisherPresent: isPublisherVisible)
+        }
+        
+        notifyResolutionsChanged(resolutions: pendingPublisherResolutions.drain())
+    }
+    
+    mutating func notifyEnhancedLocationUpdated(locationUpdate: LocationUpdate) {
+        enhancedLocation = locationUpdate
+        delegate?.subscriber(sender: ablySubscriber, didUpdateEnhancedLocation: locationUpdate)
```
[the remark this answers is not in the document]
lawrence-forooghian: I'd suggest that these methods that need to perform delegate-related logging should log to a parameter `logHandler: InternalLogHandler?` so that these messages get logged in the context of a particular worker. So their API would look something like [this](https://github.com/ably/ably-asset-tracking-swift/commit/f5566b61f35247aea9b9ed0ffbd46c9e710ee141) (just a temporary commit on my branch; would be best if it were done in this PR) and workers would use them like [this](https://github.com/ably/ably-asset-tracking-swift/commit/2aa095269f078af0620c82683ad0ff7f24149ae1#diff-a2ad5f62ece72ca1721999eeb948e58abd7dbd613048747077f8417a16d714beR16-R19).

Sources/AblyAssetTrackingSubscriber/WorkerQueue/SubscriberWorkerQueueProperties.swift
```diff
@@ -3,4 +3,185 @@ import Foundation
...
+                guard let subscriber = subscriber
+                else { return }
+                delegate?.subscriber(sender: subscriber, didUpdateResolution: resolution)
+                delegate?.subscriber(sender: subscriber, didUpdateDesiredInterval: resolution.desiredInterval)
+            }
+        }
+    }
+}
+
+fileprivate class PendingResolutions {
```
lawrence-forooghian, line 175: Should this not be a struct, to preserve the value semantics of `SubscriberWorkerQueueProperties`?

Sources/AblyAssetTrackingPublisher/State/LocationsPublishingState.swift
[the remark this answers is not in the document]
JakubJankowski: Yeah, I reverted them.

Sources/AblyAssetTrackingSubscriber/WorkerQueue/SubscriberWorkerQueueProperties.swift
```diff
@@ -1,7 +1,181 @@
...
+    private var pendingPublisherResolutions: PendingResolutions = PendingResolutions()
+    
+    var enhancedLocation: LocationUpdate?
+    var rawLocation: LocationUpdate?
+    var trackableState: ConnectionState = .offline
+    var publisherPresence: Bool = false
+    var resolution: Resolution?
+    var nextLocationUpdateInterval: Double?
+    
+    weak var delegate: SubscriberDelegate?
```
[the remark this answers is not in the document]
JakubJankowski: Probably a good idea. 521b33c221f06ad09bcd7c116389a63d43ad1144

Sources/AblyAssetTrackingSubscriber/WorkerQueue/SubscriberWorkerQueueProperties.swift
JakubJankowski: 9c745748bba921ec848c6eb6225dd9ac5de55579

Sources/AblyAssetTrackingSubscriber/WorkerQueue/SubscriberWorkerQueueProperties.swift
JakubJankowski: Good suggestions, thanks! 👍  846237855e2d882a68495a0d63106d3d3939e815

Sources/AblyAssetTrackingCore/Model/ConnectionState.swift
JakubJankowski: I've removed it as a part of a bigger merge commit: e2b1526d379b3287c70a790c4ad457b66a73fedf

Sources/AblyAssetTrackingSubscriber/WorkerQueue/SubscriberWorkerQueueProperties.swift
JakubJankowski: I'm not sure it would actually cause problems in our case, but sure, it can and probably should actually be just a struct. a1f35512729b12de1185bb4762a56d31488296c1

Sources/AblyAssetTrackingSubscriber/WorkerQueue/SubscriberWorkerQueueProperties.swift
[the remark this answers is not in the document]
lawrence-forooghian: It seems you've addressed the logging but not the dispatching to the main thread; please can you add that too.

I’d suggest keeping all of this logging / dispatching logic in a single place, same as the `DefaultSubscriber.callback(event: DelegateEvent)` method I linked to above. Easier to keep it consistent that way, what do you think?

Sources/AblyAssetTrackingSubscriber/DefaultSubscriber.swift
```diff
@@ -54,6 +53,7 @@ class DefaultSubscriber: Subscriber {
         self.ablySubscriber.subscriberDelegate = self
 
         self.ablySubscriber.subscribeForAblyStateChange()
+        self.workerQueue.properties.subscriber = self
```
lawrence-forooghian, line 56: As mentioned elsewhere, we can't make the worker queue’s properties public. We need to solve this in some other way.

How about the following:

1. make `workerQueue` an implicitly unwrapped optional, and then do the following two things inside `init`:
2. initialize the local var `properties` once `self` is available in `init`, passing it `subscriber: self`
3. then set up `self.workerQueue`

JakubJankowski: A good suggestion, thanks. Perhaps not a biggest fan of implicitly force unwrapping in the property declaration, but in this case I'd say it's acceptable, considering the alternatives.
f0266d7194f0322a800dbafa0f31c35d2156fc63

Sources/AblyAssetTrackingInternal/WorkerQueue/WorkerQueue.swift
```diff
@@ -14,7 +14,8 @@ import Foundation
 ///     semantics
 ///     - WorkerSpecificationType - the type of specification used to post worker back to the queue
 public class WorkerQueue<PropertiesType, WorkerSpecificationType> where PropertiesType: WorkerQueueProperties {
-    var properties: PropertiesType
+    // swiftlint:disable:next missing_docs
+    public var properties: PropertiesType
```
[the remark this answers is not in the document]
JakubJankowski: Had to expose it to set the `workerQueue.properties.subscriber` in the `DefaultSubscriber`'s init: 
```
        self.workerQueue.properties.subscriber = self
```
as you suggested elsewhere in this PR. 

Removed as per your suggestion below in this commit: f0266d7194f0322a800dbafa0f31c35d2156fc63

Sources/AblyAssetTrackingSubscriber/WorkerQueue/SubscriberWorkerQueueProperties.swift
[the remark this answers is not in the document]
JakubJankowski: 46713fbe215964c99ad417d7b83e3f64b993602c

Sources/AblyAssetTrackingSubscriber/DefaultSubscriber.swift
lawrence-forooghian: Yeah, this is the kind of situation that implicitly unwrapped optionals are intended for, I think — when you have two classes with references to each other and where one wants to create the other as part of its initialization.

lawrence-forooghian (approved):
Looks good 👍  I've also resolved the merge conflicts.

PR #629, opened 2023-03-21
Set up SwiftLint rules

### LINKED ISSUE #628: Revisit SwiftLint rules
In #627 we re-enabled SwiftLint but with lots of rules disabled in order to get it passing. We should revisit all of our SwiftLint rules, decide which ones we want enabled, and fix violations.

lawrence-forooghian:
This sets up what I think is a sensible SwiftLint config. See commit messages for more details.

Resolves #628.


Tests/InternalTests/PresenceDataTests.swift
```diff
@@ -26,11 +26,13 @@ class PresenceDataTests: XCTestCase {
         jsonString = "{\"type\":\"PUBLISHER\",\"rawLocations\":true}"
         data = try PresenceData.fromJSONString(jsonString)
         XCTAssertEqual(data.type, .publisher)
+        // swiftlint:disable:next xct_specific_matcher
```
JakubJankowski, line 29: Any specific reason why we're ignoring this rule here, instead of using `XCTAssertTrue`, and `XCTAssertFalse` below?

lawrence-forooghian: Yep — it's because `data.rawLocation` is a `Bool?` and hence can't be used with `XCTAssertTrue/False`, but SwiftLint doesn’t realise this.

JakubJankowski (approved):
In general LGTM, didn't find any rules I'd consider unreasonable. Only added a single minor comment.

Tests/InternalTests/PresenceDataTests.swift
lawrence-forooghian: Good idea, changed it :)

PR #644, opened 2023-04-13
Improve README, especially Mapbox auth instructions

lawrence-forooghian:
See commit messages for more details.


lawrence-forooghian:
> Additionally, I know it's not the scope of this PR, but "Configure credentials" for map box [here](https://github.com/ably/ably-asset-tracking-swift#example-apps) should be also changed.

Thanks! I've replaced those instructions with a pointer to the earlier ones.

